PR #339, opened 2026-05-28
Fix ejamit shapefile no-block polygon area assignment

### LINKED ISSUE #338: fix crash in ejamit()  when polygon has zero blockpoints inside it (maybe only in ACS2024 branch aka v2.5.0 draft)
this fails because the code that adds area_sqmi column is not handling the fact that the shapefile has 2 valid rows (polygons) and 1 invalid row (not valid, in the sense of having no reported results because no blockpoints were found inside that polygon)
```
shp = shape_buffered_from_shapefile_points(  shapefile_from_sitepoints( testpoints_100[23:25,]), radius.miles = 1)
x = ejamit(shapefile = shp)

Error in set(x, j = name, value = value) : 
  Supplied 3 items to be assigned to 2 items of column 'area_sqmi'. 
  If you wish to 'recycle' the RHS please use rep() to make this intent clear to readers of your code.
4. set(x, j = name, value = value)
3. out$results_bysite$area_sqmi <- areas at ejamit.R#715
2. out$results_bysite$area_sqmi <- areas at ejamit.R#715
1. ejamit(shapefile = shp)
```

ejanalysis:
## Summary

- Fixes #338 by assigning shapefile area_sqmi values by ejam_uniq_id instead of assuming every valid polygon has a returned aggregation row.
- Adds a regression test using the issue's three buffered polygons, where one polygon has no blockpoints and should be returned as invalid instead of crashing.

## Validation

- Reproduced the original crash on development before the fix.
- Verified the issue example now returns three results_bysite rows with one invalid no-block polygon.
- devtools::test(filter = area_sqmi) passed: 39 OK, 0 failures, 0 warnings, 0 skips.

## Note

A broader devtools::test(filter = ejamit) run still has unrelated existing fixture/data differences on development; the new focused regression is in tests/testthat/test-area_sqmi.R to avoid conflating those with this narrow crash fix.


ejanalysis:
closing #339 as superseded by the narrowly focused #340

PR #383, opened 2026-06-02
Make v3.2022 the default: merge ACS2022 (v3.2022.0, ACS 2018-2022) into main

ejanalysis:
Makes **v3.2022.0** (ACS 2018-2022) the package default — an improvement over v2.32.8.001 that keeps the same ACS vintage but adds:
- the 6/1 web-app/reporting/install fixes (PR #377 + #364), now merged into the v3 line;
- newer FRS facility data and app enhancements (PDF reports, barplot/Details improvements, ejamapi);
- Island-Areas support and a modernized dataset build (full EJSCREEN-style `blockgroupstats` schema, `DWATER` missing-as-NA, `bgid` as double);
- the versioned-pkgdown workflow + `docs/` removal (this **supersedes #381/#382**).

`origin/main` is now an ancestor of `ACS2022` (main's 6/1 fixes were merged into the v3 line first), so this is a clean merge.


ejanalysis:
One concrete blocker here: this branch drops several existing exported entry points from `NAMESPACE` even though the functions still exist, which is a breaking API regression for current callers. For example, `main` exports `app_run_EJAM`, `ejamapi_local`, `calc_blockgroupstats_acs`, `calc_blockgroupstats_from_tract_data`, `calc_formulas_from_varname`, `calc_pctile_columns`, and `calc_ratio_columns` (`NAMESPACE` lines 8, 16-21, 61), but the PR branch no longer exports them (`ACS2022:NAMESPACE` lines 3-16, 54-55). `run_app()` / `app_run_EJAM()` are also still defined in `R/ejamapp.R` lines 387-409, so this looks accidental rather than an intentional removal. Unless this major-version bump is explicitly meant to break those APIs, I think these exports need to stay or there should be a deliberate compatibility/deprecation path.

ejanalysis:
Heads-up: this branch drops several currently exported entry points from the package namespace while keeping the implementations in-tree (for example `run_app()` / `app_run_EJAM()` in `R/ejamapp.R`, `ejamapi_local()` in `R/ejamapi_local.R`, `getblocks_predict_blocks_per_site()` / `getblocks_summarize_blocks_per_site()` in `R/getblocks_diagnostics.R`, and `calc_blockgroupstats_acs()` / `calc_blockgroupstats_from_tract_data()` in their existing source files). As written, installing v3.2022.0 from this PR would turn those names into non-exported objects and break downstream code that currently calls `EJAM::...` for them. If that API contraction is intentional, it needs an explicit migration/deprecation plan; otherwise the exports should stay.

ejanalysis:
I found one runtime wiring issue in the shared ACS2022 head. In R/calc_ejscreen_blockgroupstats.R lines 193-197, reuse_existing_extra_if_missing is forwarded into complete_bg_geodata() as reuse_existing_if_missing. That means turning on fallback for missing extra indicators also turns on the geodata fallback path, which can reuse stale arealand/areawater from existing_blockgroupstats or hard-stop on subset runs because R/calc_bg_geodata.R lines 208-213 require an exact bgfips universe match unless allow_partial_reuse=TRUE is exposed. I do not see a test that covers this real call path; tests/testthat/test-calc_ejscreen_dataset.R lines 57-78 only stub the helper.

ejanalysis:
Two release-risk issues stood out in review:

1. `DESCRIPTION` moves `ACSdownload` to `Suggests`, but production entry points still reference it directly. In particular, `calc_ejscreen_dataset()` now defaults `acs_download_fun = ACSdownload::get_acs_new` (`R/calc_ejscreen_dataset.R`), and other runtime paths call `ACSdownload::get_acs_new()` directly. A standard install without suggested packages will break those code paths immediately.

2. `NAMESPACE` drops a large set of existing exports from the current release line (`acs_table_info`, `calc_blockgroupstats_acs`, `calc_blockgroupstats_from_tract_data`, `ejamapi_local`, `run_app`, etc.). That is a user-visible API break for downstream callers, and I did not see namespace/API contract coverage that would make this safe.

ejanalysis:
Review note: this PR removes several existing public exports from the package namespace even though the functions still exist in the branch (for example run_app(), app_run_EJAM(), acs_table_info(), ejamapi_local(), calc_blockgroupstats_acs(), calc_blockgroupstats_from_tract_data(), calc_formulas_from_varname(), calc_pctile_columns(), and calc_ratio_columns()). In development, these are exported today, but in the PR head they are no longer exported from NAMESPACE. That is a breaking API change for existing callers such as EJAM::run_app(), and I do not see coverage that protects the export contract; the tests mostly exercise these via EJAM:::. If the removals are intentional, this likely needs an explicit deprecation/migration plan and release-note callout before merging to main.

ejanalysis:
I found one release-risk issue in `R/utils_acs_endyear.R`: `acs_endyr_latest_published_by_ejscreen()` hard-codes `2024` for every `guess_as_of >= "2026-06-01"` (lines 72-76), but this PR is making `v3.2022.0` / ACS 2022 the default release. After merge, `acs_endyear(guess_always = TRUE)` would therefore report ACS 2024 for dates like 2026-06-02 even though the package being released here is still ACS 2022. The current test coverage only checks 2026-02-01 -> 2022 in `tests/testthat/test-acs_endyear.R` and never exercises the June-2026 branch, so this mismatch would slip through. I would gate that 2026-06-01 override on the actual package release state, or keep returning 2022 here until the ACS 2024 line is actually the default package release.

ejanalysis:
Two release-level issues stood out in the non-doc changes:

1. NAMESPACE drops a large set of previously exported user-facing functions (run_app, app_run_EJAM, ejamapi_local, calc_blockgroupstats_acs, calc_blockgroupstats_from_tract_data, calc_formulas_from_varname, calc_pctile_columns, calc_ratio_columns, custom_doaggregate, proxistat, etc.). That is an immediate runtime break for downstream code using EJAM::..., and I did not see compatibility wrappers or namespace-compat tests for the removed exports.

2. The new pipeline entrypoints now default to ACSdownload::get_acs_new (calc_ejscreen_dataset() and download_bg_acs_raw()), but ACSdownload moved to Suggests plus a GitHub Remotes entry in DESCRIPTION. On a standard install without Suggests or GitHub access, the primary pipeline path will fail at runtime as soon as it needs raw ACS downloads. This looks release-blocking unless the dependency is imported or the high-level pipeline degrades cleanly with explicit install handling and coverage for the missing-package case.

Issue #342, opened 2026-05-28
use renv package to create lock state for what versions of what packages are used


ejanalysis:
The EJAM repo’s dependency state is defined mainly by `EJAM/DESCRIPTION` and `EJAM/data-raw/Dockerfile`.
There is no checked-in lockfile, and the Docker image uses rocker/rstudio:latest plus mostly unversioned install2.r installs. That means the repo does not record the actual resolved R package set, only minimum version floors.

We should add a lock state (renv.lock or equivalent). That turns future dependency sweeps into real versioned upgrades instead of guesswork.

This issue potentially affects web app if obsolete dependency is not caught or there is a problem trying to install the latest EJAM release for hosting or API purposes.

As of May, 2026:

For EJAM or a similar R package, recommendation is to use: 

- **renv.lock for reproducible development/ CI/ release environments** since [renv](https://rstudio.github.io/renv) is the mainstream recommended solution for reproducible R package-development environments. "R Packages" guidance explicitly points to renv when you need exact-version freezing beyond DESCRIPTION ([R Packages, DESCRIPTION](https://r-pkgs.org/description.html)).
- **pak for installing/resolving dependencies**, since the [pak package](https://pak.r-lib.org/reference/pak.html) is a new and useful complement to renv. pak is already used in some places in EJAM.
- DESCRIPTION + NAMESPACE for package dependency truth

**renv + pak is a good current combo**: renv owns the project lockfile; pak can be used as the installation backend. Current renv has official pak support, and recent renv releases continue improving that integration ([pak manual](https://r-lib.r-universe.dev/pak/doc/manual.html), [renv changelog](https://rstudio.github.io/renv/news/index.html)).

**renv** is the lockfile/environment tool: it creates a project library and records exact package versions in renv.lock, then restores them later with renv::restore() ([renv overview](https://pkgs.rstudio.com/renv/index.html), [lockfile docs](https://rstudio.github.io/renv/articles/lockfile.html)).

**pak** is the modern installer/resolver: use it for fast, robust installs, GitHub/remotes, system dependency detection, and dev dependencies. It can install all development dependencies for the current package with pak::pak() when pkg = NULL ([pak docs](https://pak.r-lib.org/reference/pak.html)).

packrat was an older obsolete package superseded by renv.

renv does record the R version but does not install R for you. If we also need to pin/manage the R executable itself, we should add rig or containerization.

ejanalysis:
closely related to #167

[...]

PR #416, opened 2026-06-28
[codex] Prepare dependency floor updates

ejanalysis:
## Summary

- Raise dependency floors for `arrow`, `curl`, and `shiny`, with CI/runtime guards for Arrow/curl and Shiny 1.14 download-button behavior.
- Update EJAM code for ggplot2 4.x by replacing the deprecated `stat_summary(fun.y = ...)` usage.
- Add focused regression tests and update `R/test_ejam.R` group/timing metadata for the new test files.

## Why

This keeps the dependency sweep narrow and ties each included update to a concrete EJAM use case:

- `shiny >= 1.14.0`: Shiny 1.14 adds `enabled` support to `downloadButton()` / `downloadLink()`, with automatic download-readiness behavior by default. EJAM already has explicit `shinyjs::disable()` / `shinyjs::enable()` readiness logic for the multisite report and results-table downloads, so this PR opts those buttons out with `enabled = FALSE` and then tests that contract. Shiny 1.14 also improves output resize/visibility handling and `conditionalPanel()` startup behavior, both relevant to EJAM's tabbed UI, plots, maps, and conditional controls.
- `arrow >= 24.0.0`: EJAM relies on Arrow-backed large datasets and pipeline stage I/O. Requiring the current tested Arrow release lets CI validate the same Arrow IPC read/write path EJAM uses for pipeline files and reduces drift between the declared package floor and the Arrow version used in current package-data/pipeline workflows.
- `curl >= 7.1.0`: EJAM installation, release, and data workflows depend on modern libcurl-backed download behavior. This PR pairs the R package floor with CI checks for the installed `curl` package and `libcurl >= 7.73.0`, so platform/toolchain problems surface in CI instead of failing later during user installs or data downloads.
- ggplot2 4.x compatibility: this PR does not raise the `ggplot2` floor yet, but it removes EJAM's legacy `stat_summary(fun.y = ...)` call and uses the current `fun = mean` argument. That makes the plotting code ready for ggplot2 4.x before any future floor bump.
- `golem` and `jsonlite` remain intentionally deferred because they are larger major-version floor changes and need their own targeted migration/testing PRs.

## Validation

- `source(".github/scripts/check-arrow-curl-deps.R")` with `arrow 24.0.0`, `curl 7.1.0`, `libcurl 8.14.1`
- `testthat::test_file("tests/testthat/test-shiny-1-14-compat.R")` under Shiny 1.14.0
- `testthat::test_file("tests/testthat/test-ejam2boxplot_ratios.R")` under ggplot2 4.0.3
- `testthat::test_file("tests/testthat/test-ejscreen-pipeline-io.R")`
- workflow YAML parse for 6 `.github/workflows/*.yaml` files
- `R CMD INSTALL --library=/private/tmp/ejam-shiny-1.14-lib .`
- `shinytest2::test_app(".", filter = "FIPS-picker-functionality", check_setup = FALSE)` with 8 pass / 0 fail / 0 warn / 0 skip
- `EJAM:::test_ejam(...)` narrow validation path for file-list/timing metadata, running only the `test` group
- `git diff --cached --check`


ejanalysis:
somewhat related to #342 indirectly

PR #435, opened 2026-07-01
Fix ECS prod deploy: use github.token instead of expired EJAMDATA_PAT

ejanalysis:
## Problem
Same expired-PAT issue as dev (fixed in #434). The prod deploy workflow builds with `--build-arg GITHUB_PAT=${{ secrets.EJAMDATA_PAT }}`, a stored PAT that has expired → `install_github` gets HTTP 401 'Bad credentials', breaking even the public-repo installs.

## Fix
Switch both deploy workflows to `${{ github.token }}` (auto-generated, never expires), matching every passing R CI workflow. Both `EJAM` and `ejamdata` are public, so a repo-scoped token has sufficient read access.

## ⚠️ Deployment note
Merging this to `prod-deploy` triggers a **live production** ECS Fargate deployment. Merge deliberately when you're ready to redeploy prod.

### DIFF
--- a/.github/workflows/deploy-dev.yaml
+++ b/.github/workflows/deploy-dev.yaml
@@ -47,11 +47,11 @@ jobs:
         run: |
           FULL_IMAGE=$REGISTRY/$ECR_REPOSITORY
 
-          # Use the auto-generated, always-valid GITHUB_TOKEN (mirrors the R CI
+          # Use the auto-generated GITHUB_TOKEN for this workflow run (mirrors the R CI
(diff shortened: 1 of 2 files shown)


ejanalysis:
Addressed the Copilot review wording nit in **both** deploy workflows. The `deploy-dev.yaml` correction was already in `2a2f4c6b`; commit `9e8f9759` applies the identical wording to `deploy.yaml`, so both files now describe `GITHUB_TOKEN` accurately (minted fresh per run / short-lived, not "always-valid"). Both review threads are resolved.

PR #431, opened 2026-07-01
Deploy Dockerfile: pin EJAM via install_github(@EJAM_VERSION) instead of install_local of the build context

ejanalysis:
Addresses the `remotes::install_local('/home/epic/')` concern in the deploy-branch Dockerfile.

### What it was actually doing
`/home/epic/` was **not** a reference to anyone's local machine — the Dockerfile did `mkdir -p /home/epic`, then `ADD . /home/epic/` (copy the **build context** = this deploy branch's checkout), then `install_local('/home/epic/')`. So it installed **whatever EJAM source is on the deploy branch**. That branch is currently the **stale 2.32.8.001 snapshot**, so a deploy today would install 2.32.8.001 — the version was implicit and easy to get wrong.

### The fix
- **`ADD .`+`install_local` → `remotes::install_github('Public-Environmental-Data-Partners/EJAM@' + EJAM_VERSION)`**, pinned via a new **`ARG EJAM_VERSION=v3.2022.1`** build-arg (same mechanism as the EJAM-API image). EJAM is a **public** repo and its `data/*.rda` ship in the release tarball, so **no token and no Git-LFS** are needed (LFS only covers dev-only `data-raw/` files).
- Removed the `/home/epic` scaffolding (`WORKDIR /root`).
- The container entrypoint runs the **installed** `EJAM::run_app()`, so nothing else needed the copied build context.

### Why it's better
Explicit, reproducible, version-pinned deploys, decoupled from deploy-branch drift, and consistent with the API. To ship a different version, set `--build-arg EJAM_VERSION=vX.Y.Z` (or change the default) — no branch-code sync required.

### ⚠️ Semantic change to be aware of
The deployed EJAM version is now the **`EJAM_VERSION` build-arg (default v3.2022.1)**, not the branch's checked-out source. The `deploy-dev.yaml` workflow currently passes only `--build-arg GITHUB_PAT=…`, so the **default (v3.2022.1) governs**. If you later want the workflow to choose the version, add `--build-arg EJAM_VERSION=…` there.

### Related follow-up (not in this PR)
The ejamdata download still defaults `EJAMDATA_VERSION` to the **latest** ejamdata release; for a matched v3.2022.1 build it should be pinned to **v3.2022.0** (EJAM v3.2022.1's `ejamdata_required_tag`). Happy to add that (either a `v3.2022.0` default or derive it from the installed EJAM's DESCRIPTION).

### Testing / merge note
[...]

PR #438, opened 2026-07-01
infra: ECS health-check grace period 0 → 300s (suspected fix for dev v3.2022.1 stabilization failure)

ejanalysis:
## Issue
The v3.2022.1 dev deploy was failing but now works. While looking at it, noted that the ECS service has `healthCheckGracePeriodSeconds: 0`, and the ALB health check (port 2001, path /, 30s interval × 3 unhealthy) gives a task only **~90 seconds** before it is flagged. The container CMD starts the port-2001 httpuv health endpoint **in the same R process** as the Shiny app, so it cannot answer until R finishes `library(EJAM)` + loading the arrow data — on the 1-vCPU dev task, plausibly longer than 90s for the new (Public-Environmental-Data-Partners/EJAM#431) image.

## Change
Add `health_check_grace_period_seconds` (variable, default **300**) to `aws_ecs_service.app`. Grace only suppresses *health-check-based* kills during startup — a genuinely crashed container still exits and stops the task immediately, so this cannot mask a real crash-loop.

## How to apply (ORDER MATTERS — the deploy workflow does not run terraform)
1. Apply the infra change first, either:
   - `terraform apply -var-file=dev.tfvars` from `ejam-infra/` on this branch, **or**
   - the zero-terraform equivalent: `aws ecs update-service --cluster ejam-dev-cluster --service ejam-dev-service --health-check-grace-period-seconds 300` (then merge this PR so IaC matches and a future terraform apply does not revert it).
2. Re-run the dev deploy ("Build & Deploy to ECS Fargate (dev)" has workflow_dispatch) and watch `services-stable`.
3. If it stabilizes → hypothesis confirmed, dev serves v3.2022.1. If it still fails → it is a crash inside the image; grab `aws logs tail /ecs/ejam-dev --since 1h`.

`main.tf` is identical on both deploy branches, so port this to `prod-deploy` together with the Public-Environmental-Data-Partners/EJAM#431 Dockerfile changes when promoting v3.2022.1 to prod.

Context: dev pipeline itself is proven good (build green after Public-Environmental-Data-Partners/EJAM#434; prod ran the same workflow end-to-end successfully 2026-07-01 with the old image via Public-Environmental-Data-Partners/EJAM#435).

### DIFF
--- a/ejam-infra/main.tf
+++ b/ejam-infra/main.tf
@@ -89,6 +89,11 @@ variable "health_check_port" {
   default = 2001
 }
 
+variable "health_check_grace_period_seconds" {
+  description = "How long ECS ignores ALB health-check failures after a task starts. The container CMD starts the port-2001 health endpoint in the same R process as the Shiny app, so it cannot answer until R finishes loading EJAM + arrow data; with the default of 0, the ALB (30s interval x 3 unhealthy = ~90s) kills slow-booting tasks before they are ready. Grace only suppresses health-check kills during startup - a crashed container still stops immediately."
+  default     = 300
+}
+
 variable "task_cpu" {
   description = "Fargate task CPU units (1024 = 1 vCPU)"
   default     = 2048
@@ -509,6 +514,8 @@ resource "aws_ecs_service" "app" {
   task_definition = aws_ecs_task_definition.app.arn
   desired_count   = var.desired_count
   launch_type     = "FARGATE"
+
+  health_check_grace_period_seconds = var.health_check_grace_period_seconds
 
   network_configuration {
     subnets          = aws_subnet.public[*].id


ejanalysis:
**Verdict update — grace-period theory REFUTED.** The read-only diagnostics (run https://github.com/Public-Environmental-Data-Partners/EJAM/actions/runs/28551072365) show tasks dying with `stopCode: EssentialContainerExited`, exit 1 — an app crash, not an ALB health-check kill: `Error: 'run_app' is not an exported object from 'namespace:EJAM'`. Root cause fixed in Public-Environmental-Data-Partners/EJAM#440 (CMD now calls the exported `ejamapp()`).

Keeping this PR open as **optional hardening only** (grace=0 remains a latent footgun if boot ever slows), but it is NOT needed to unblock the v3.2022.1 rollout. Recommend deciding after Public-Environmental-Data-Partners/EJAM#440 proves out on dev.

Issue #49, opened 2025-09-11
URLs / links to EJScreen reports for FIPS analysis other than counties - need these to work


carlyhagen:
---ISSUE FROM ARCHIVE---
Originally created: 11/15/2023
-----------------

IMPORTANT: as this issue has be reuploaded from an archived repository, some of the links may be broken or no longer relevant. Critical resources can be found through https://ejanalysis.com/

ejanalysis:
Enhancement triage on ACS2024: closing as implemented for the current EJSCREEN map-link behavior. `url_ejscreenmap(fips = ...)` now returns EJSCREEN map URLs for tested blockgroup, tract, city/CDP, county, and state FIPS examples by converting FIPS inputs to usable map locations.

ejanalysis:
#49 was fixed in EJAM v2.32.6, GitHub release “EJAM v2.32.6 (September 2025)”, tag v2.32.6, published 2025-09-29.

The key release commit was 5525665cde220ba0724649c0229eb6f973ec78f2 (EJAM version 2.32.6 (September 2025) release (#108)).

Issue #48, opened 2025-09-11
URLs / links to EJScreen reports for shapefile analysis


carlyhagen:
---ISSUE FROM ARCHIVE---
Originally created: 11/15/2023
-----------------

In excel and in interactive datatables, EJScreen report URL using their API for a report needs to work for shapefile-based analysis/ polygon(s) not just circular buffers or County FIPS reports. 

### Quick interim fix needed asap

Immediate interim workaround should be to replace those links in all tables with NA (or just link to map not report), when the analysis uses shapefiles!  

### Longterm fix

When you run a shapefile report in EJAM, such as using the testdata "NJ SEWERSHEDS WITH NEEDS" files in "testdata/shapes/NJ Sewersheds/", the excel spreadsheet of site by site results tries to show the usual column of links to EJScreen reports, but those links as coded do not work for a report on a polygon - the function creating those URLs was designed only to handle a circular buffer report (or a county FIPS also works) in EJScreen. The EJScreen API probably can handle a polygon input but likely must be a more complicated query like POST instead of GET and providing geometry of polygon in query body? Not sure simple URL-encoded query would make sense for that case.  

### Closely related issue - other kinds of FIPS?

As a closely related issue, the EJScreen API may only support fips that are for tract or county, but not State, Place, CBSA, EPA Region, etc., so EJAM should handle those types of FIPS-based analyses by converting those FIPS Census units into polygons/boundaries, and then asking the EJScreen API service for a report on the polygon that is the boundaries of each such Census unit. However, if this is unrealistic to query that service for such large polygons as whole States, then consider not providing that option for units/polygons above some size or of certain categories.

---ARCHIVED COMMENTS---

Urgency high for the "quick fix" but medium or low for the longterm fix. |Also see #652 

IMPORTANT: as this issue has be reuploaded from an archived repository, some of the links may be broken or no longer relevant. Critical resources can be found through https://ejanalysis.com/

ejanalysis:
Last ranked: 2026-05-30
Cost score: 8
Benefit score: 11
Quadrant: B
Run report: inst/issues/ejam_issues_scored_by_risk_and_value.MD

ejanalysis:
Cross-reference: #349 is closely related because it covers polygon report links and result mismatches in the same report-link/API path.

ejanalysis:
[...]

PR #468, opened 2026-07-16
Fix reports for zero-population sites

### LINKED ISSUE #467: ejam2report() fails for zero-population or otherwise analysis-invalid sites
## Summary

`ejam2report()` can return a bare `NA` instead of creating a report when an `ejamit()` result contains a site with `valid = FALSE` because no usable population results were produced (for example, no block centroids in the radius, zero residents, or a site dropped during aggregation).

This affects direct R use and the report endpoint in `Public-Environmental-Data-Partners/EJAM-API#48`: a bare `NA` can become a corrupt 1-byte response labeled as a PDF. See the related review: https://github.com/Public-Environmental-Data-Partners/EJAM-API/pull/48#pullrequestreview-4710426796

## Reproduction

```r
library(EJAM)

x <- ejamit(lat = 33, lon = -112, radius = 1.1)
ejam2report(x) # OK

x <- ejamit(lat = 33, lon = -112, radius = 1)
ejam2report(x) # can fail / explicit sitenumber = 1 returns NA

x <- ejamit(
  lat = c(33, 33.00001),
  lon = c(-112, -112.01),
  radius = 1
)
ejam2report(x) # returns NA when row 1 is invalid and row 2 is valid
ejam2report(x, sitenumber = 2) # OK
```

## Root cause

There are three related contract/selection problems:

1. `doaggregate()` intentionally omits sites that it cannot aggregate. `ejamit()` later merges the original sites back into `results_bysite`, and those reinserted rows currently acquire `pop = NA`. For final `ejamit()` site results, `valid = FALSE` already carries the availability distinction, so population should be `0` for these rows while the unavailable indicator values remain `NA`.
2. When exactly one row has `valid = TRUE`, `ejam2report()` changes an omitted `sitenumber` to `1`. It assumes the only valid result is row 1 instead of selecting the row whose `valid` value is actually true. A mixed result with invalid row 1 and valid row 2 therefore selects the invalid row and returns `NA`.
3. For an explicitly selected row with `valid = FALSE`, `ejam2report()` rejects the row and returns `NA`, even when the input location was usable and the analysis simply found no population to summarize.

The third path matters directly to EJAM-API, which passes an explicit site number for single-site reports.

## Historical behavior

This is a long-standing latent bug rather than a new `ejam2report()` regression. The same row-1 assumption and `valid = FALSE` -> `return(NA)` gate are present in tag `v2.32.8.001`. Running current example result objects through that tagged report code reproduces `NA` for both the explicit zero-population site and the mixed-site default.

Changes in datasets, inputs, or API use can expose the old behavior more often, but the failing report logic itself was already present in `v2.32.8.001`.

## Expected behavior

- In final `ejamit()` output, `results_bysite$pop` is `0` for rows with `valid = FALSE`; other unavailable result columns remain `NA`.
- `results_overall$pop` retains the aggregate population calculated by `doaggregate()`; adding invalid site rows with population zero must not alter it or cause double-counting across overlapping sites.
- A usable location with zero population or no analysis results still produces the normal EJAM report, showing `0`/`NA` values as appropriate.
- If automatic one-valid-site behavior is retained, it selects the actual valid row rather than row 1.
- Truly malformed inputs retain an explicit not-reportable path rather than failing while building a map.

## Acceptance criteria

- Regression coverage for `ejamit()` invalid-site population normalization.
- Regression coverage for a selected zero-population/no-results site.
- Regression coverage for a mixed result where row 1 is invalid and row 2 is valid.
- A non-empty normal report is returned for reportable no-results sites, with population `0` and other unavailable values preserved as `NA`.
- Existing behavior for positive-population reports remains unchanged.
- The fix is compatible with the EJAM-API report path described in `Public-Environmental-Data-Partners/EJAM-API#48`.

— Codex

ejanalysis:
## Summary

- normalize final `ejamit()$results_bysite$pop` to `0` for `valid = FALSE` rows while preserving `valid = FALSE` and the other unavailable values as `NA`
- keep historical batch-summary handling unchanged by applying that normalization only to final site-level output
- make `ejam2report()` select the actual valid row when only one of several rows is valid
- allow normal reports for known analysis-no-results cases with population zero, while retaining the not-reportable path for malformed inputs
- centralize the reportable analysis-invalid messages so emitter wording and reportability checks cannot drift apart

## Root cause

`doaggregate()` intentionally omits sites it cannot aggregate. `ejamit()` merges the original input rows back afterward, which previously left population missing. Separately, `ejam2report()` assumed that a sole valid site was row 1 and rejected every explicitly selected `valid = FALSE` row with a bare `NA`.

The same report selection and validity-gate behavior is present in `v2.32.8.001`, so this is a long-standing latent bug rather than a newly introduced report regression. Dataset and API usage changes expose it more readily.

The fix does not recompute `results_overall$pop` from site rows; that aggregate remains the union-aware result from `doaggregate()`, avoiding double-counting across overlapping sites.

## Impact

Direct R calls can now create ordinary HTML/PDF reports for usable locations with zero population or no analysis results. Population displays as zero, unavailable indicators remain `NA`, and malformed locations are still rejected.

EJAM-API will be able to serve these reports after it updates to an EJAM release containing this fix and relaxes the pre-report zero-population guard in Public-Environmental-Data-Partners/EJAM-API#48. Its post-render `NULL`/`NA` guard should remain.

Related API review: https://github.com/Public-Environmental-Data-Partners/EJAM-API/pull/48#pullrequestreview-4710426796

## Validation

- `devtools::test(filter = "ejam2report")`: PASS 37, FAIL 0, WARN 0, SKIP 1 (Pandoc unavailable for one pre-existing path test)
- exact one-site and mixed-site examples: both generated normal 2.8 MB HTML reports; invalid site population was `0`; the mixed case selected row 2
- real zero-population PDF: six pages, 605,120 bytes
- review follow-up: all 63 `ejamit` tests passed; focused shared-message checks passed; the combined affected-file run's only failure was an environment-only headless-Chrome failure in a pre-existing PDF fallback test
- `git diff --check origin/development...HEAD`
- staged-content scan found no local user paths

Closes #467

— Codex

### DIFF
--- a/R/ejam2report.R
+++ b/R/ejam2report.R
@@ -84,10 +84,7 @@ ejam2report_site_is_reportable <- function(site_result) {
   } else {
     ""
   }
-  isTRUE(grepl(
-    "^(no block centroids \\(|blocks found but zero residents$|blocks with residents found but unable to aggregate$)",
-    invalid_msg
-  ))
+  isTRUE(invalid_msg %in% unname(ejamit_reportable_invalid_messages()))
 }
 ################################################## #
 
--- a/R/ejam2report.R
+++ b/R/ejam2report.R
@@ -67,6 +67,30 @@ assert_pdf_report_available <- function() {
 }
 ################################################## #
 
+# helper
+
+ejam2report_site_is_reportable <- function(site_result) {
+  if (!("valid" %in% names(site_result)) || isTRUE(site_result$valid[1])) {
+    return(TRUE)
+  }
+
+  if (!("pop" %in% names(site_result)) ||
+      !isTRUE(!is.na(site_result$pop[1]) && site_result$pop[1] == 0)) {
+    return(FALSE)
+  }
+
+  invalid_msg <- if ("invalid_msg" %in% names(site_result)) {
+    as.character(site_result$invalid_msg[1])
+  } else {
+    ""
+  }
+  isTRUE(grepl(
+    "^(no block centroids \\(|blocks found but zero residents$|blocks with residents found but unable to aggregate$)",
+    invalid_msg
+  ))
+}
+################################################## #
+
 
 #' View HTML Report on EJAM Results (Overall or at 1 Site)
 #'
@@ -271,11 +295,12 @@ ejam2report <- function(ejamitout = testoutput_ejamit_10pts_1miles,
     sitenumber <- 0  # in case sitenumber was invalid
   }
   # How many sites were actually analyzed, in the (valid) results provided?
-  nsites <- NROW(ejamitout$results_bysite[ejamitout$results_bysite$valid %in% TRUE, ]) # might differ from ejamout1$sitecount_unique
+  valid_site_rows <- which(ejamitout$results_bysite$valid %in% TRUE)
+  nsites <- length(valid_site_rows) # might differ from ejamout1$sitecount_unique
   # Treat it like a 1-site report if only 1 valid site was analyzed.
   #   And then if sitenumber omitted, or sitenumber=1, or sitenumber provided was invalid, just use that 1 site.
   if (sitenumber %in% 0 && nsites == 1) {
-    sitenumber <- 1
+    sitenumber <- valid_site_rows[1]
   }
 
   # Multi-site  (results_overall) ###################################################
@@ -348,7 +373,7 @@ ejam2report <- function(ejamitout = testoutput_ejamit_10pts_1miles,
   include_ejindexes <- any(names_ej_pctile %in% colnames(ejamout1))
 
   if (!("valid" %in% names(ejamout1))) {ejamout1$valid <- TRUE}
-  if (isTRUE(ejamout1$valid)) {
+  if (ejam2report_site_is_reportable(ejamout1)) {
 
     #############################################################################  #
 
--- a/R/ejamit.R
+++ b/R/ejamit.R
@@ -16,6 +16,30 @@ ejamit_no_block_centroids_message <- function(sitetype) {
   "no block centroids (radius too small for low pop density)"
 }
 
+#' Normalize population for invalid sites in final `ejamit()` output
+#'
+#' `doaggregate()` omits sites it cannot aggregate. When [ejamit()] merges those
+#' original input rows back into `results_bysite`, their result columns are
+#' unavailable, but their contribution to analyzed population is zero. Preserve
+#' `NA` for the unavailable indicators while using zero for population.
+#' @param results_bysite final site-level results table from [ejamit()]
+#' @return `results_bysite` with `pop = 0` where `valid = FALSE`
+#' @noRd
+#'
+ejamit_invalid_site_pop_zero <- function(results_bysite) {
+  if (!is.data.frame(results_bysite) ||
+      !all(c("valid", "pop") %in% names(results_bysite))) {
+    return(results_bysite)
+  }
+
+  if (data.table::is.data.table(results_bysite)) {
+    results_bysite[valid %in% FALSE, pop := 0]
+  } else {
+    results_bysite$pop[results_bysite$valid %in% FALSE] <- 0
+  }
+  results_bysite
+}
+
 #' Get an EJ analysis (residential population and environmental indicators) in or near a list of locations
 #'
 #' @description This is the main function in EJAM that runs the analysis.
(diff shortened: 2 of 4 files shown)


ejanalysis (commented):
via Claude: 

Reviewed head `0a277ae5` with independent verification. **No blocking issues — this looks right and I could not break it.** Details below, including cross-PR interference checks against the three related open PRs.

## Independent verification (fresh worktree, real `ejamit()` runs — not mocks)

| Case | Result |
|---|---|
| Issue repro 1: `ejamit(lat=33, lon=-112, radius=1)` (zero-pop) | `valid=FALSE`, `pop=0`, `invalid_msg="blocks found but zero residents"`; `ejam2report(sitenumber=1)` → **2.71 MB report** (was `NA`) |
| Issue repro 3: mixed, row 1 invalid + row 2 valid, `sitenumber` omitted | selects **row 2** (report contains the valid site's lat `33.00001` and its pop ≈106; no zero-pop marker) → 2.85 MB report |
| `results_overall$pop` double-count check | overall pop = 106.5 = the valid site's pop; invalid row contributes 0, aggregate untouched ✔ |
| Regression: normal single-site + overall multisite on `testoutput_ejamit_10pts_1miles` | both render fine |
| Malformed-input gate: same zero-pop output but `invalid_msg="invalid FIPS"` | still returns `NA` (not-reportable path retained) ✔ |
| Tests: `test-ejam2report.R` + `test-ejamit.R` at this head | **99 pass / 0 fail / 0 warn / 1 skip** (the pre-existing pandoc-path skip) |

Also checked: the `invalid_msg` regex in `ejam2report_site_is_reportable()` matches the actual emitter strings (`ejamit_no_block_centroids_message()`'s three `"no block centroids ("` variants via the prefix branch, plus the two literals at their `ejamit.R` assignment sites); the "Could not capture static map snapshot" message seen during tests is pre-existing on `development`, not from this PR; no new/renamed test *files*, so no `R/test_ejam.R` group-list update is needed; no local user paths in the diff.

## Cross-PR interference check (requested)

- **Public-Environmental-Data-Partners/EJAM#466** (handoff crash fix): **no overlap** — disjoint files (`app_server.R` vs `ejamit.R`/`ejam2report.R`), both target `development`, no merge conflict in either order.
- **Public-Environmental-Data-Partners/EJAM-API#49** (radius/buffer 0 defaults for fips/shape): **complementary, no conflict** — in fact #49 makes zero-pop analyses *more likely* (buffer 0 around a tiny area), so this fix matters more with it.
- **Public-Environmental-Data-Partners/EJScreen#73**: unrelated (client-side handoff payload field).
- **Public-Environmental-Data-Partners/EJAM-API#48 — the one real coordination point.** #48's *pre*-guard 400s any single-site request where `valid=FALSE || pop<=0` **before** calling `ejam2report()`. That's correct **today** (the API pins v3.2022.1, which lacks this fix, and a bare-`NA` becomes an edge-cached 1-byte "PDF"). But once the API's EJAM pin advances to a release containing this PR, that pre-guard will **mask** the newly renderable zero-pop reports — the API would keep returning the 400 error page while the package could serve a real report. **Action for that future pin-bump: drop/relax #48's pre-guard and its README paragraph; keep its belt-and-braces post-guard (NULL/`NA` output → 500) forever.** Consequently the PR body's "EJAM-API can now create ordinary reports" is slightly ahead of reality — true only after that pin bump + guard relaxation. Leaving a note on #48 as well.

## Non-blocking notes

1. `ejam2report_site_is_reportable()` is coupled to the exact `invalid_msg` literals; any future rewording in `ejamit()` silently flips zero-pop sites back to not-reportable (fail-safe direction, but a maintenance trap). Consider centralizing the message strings as constants shared by emitter and matcher in a follow-up.
2. NEWS.md not updated — fine per the release-at-milestones policy; worth a line whenever the next release notes are assembled ("zero-population/no-results sites now get a normal report with pop 0").

ejanalysis:
Addressed the review follow-ups in `d2b6826a`:

- centralized the five reportable analysis-invalid messages in `ejamit_reportable_invalid_messages()`, which is now shared by the `ejamit()` emitters and `ejam2report()` reportability gate
- expanded focused coverage so every recognized no-results message remains reportable while malformed inputs remain excluded
- corrected the PR impact wording: direct R behavior is fixed now; EJAM-API requires a future EJAM pin update plus relaxation of Public-Environmental-Data-Partners/EJAM-API#48's pre-report zero-population guard, while its post-render `NULL`/`NA` guard should remain
- kept the NEWS item deferred to the next release-notes assembly, as suggested

Validation: parse and whitespace checks passed; all 63 `ejamit` tests passed; focused shared-message checks passed. The combined affected-file run's only failure was the environment-only headless-Chrome failure in a pre-existing PDF fallback test.

Thanks for the independent real-output verification and cross-PR check.

— Codex

Issue #446, opened 2026-07-02
Cache/precompute FIPS-based reports and boundaries (layered design: bounds, ejamit results, rendered reports)


ejanalysis:
Design proposal for caching/precomputing FIPS-based results and reports, to make community reports fast (and resilient) for the most common requests. Follows from Public-Environmental-Data-Partners/EJAM#293 (report latency) and the *Cache and/or Precalculate* section of Public-Environmental-Data-Partners/EJAM#226; closed Public-Environmental-Data-Partners/EJAM#246 applied the same idea narrowly (reusing an already-created summary for the spreadsheet download). Companion quick wins already in review: Public-Environmental-Data-Partners/EJAM#443 (links request HTML; API base routed through the Cloudflare proxy) and an EJAM-API branch that bakes the arrow data into the Docker image and makes GET /report responses edge-cacheable.

## Why FIPS specifically

FIPS units (states, counties, places, tracts) are an enumerable, deterministic request space: the same fips + same data release + same code ⇒ byte-identical results. Points/polygons are not enumerable, and **multisite combinations of 2+ FIPS are effectively not cacheable** (3,144 counties alone ⇒ ~4.9M pairs); an aggregated "overall" result requires running `ejamit()` across the union of sites. So the design caches the single-FIPS layers and makes the ingredients of the multisite path (boundaries) cheap.

## The layers cache differently

| Layer | Invalidated by | Size | Notes |
|---|---|---|---|
| **1. County boundaries baked into the package** (a `counties_shapefile` dataset + local-first branch in `shapes_from_fips()`, mirroring the existing `states_shapefile` / `shapes_state_from_statefips()` pattern) | boundary vintage (~annual, same cadence as data releases) | ~10–30 MB | Highest leverage per byte: helps every county render including multisite combos, PDF generation, buffered FIPS, and batch seeding — and removes a per-render network call to the Census/ArcGIS boundary services (a latency and failure point observed in every render). Could live in the package data or as an ejamdata release asset. |
| 2. `ejamit()` output per single FIPS | ejamdata release OR any calculation/output change | measured ~127 KB per county (full output as rds; 19 KB for just bysite+overall) | Cheap to store and survives code-only version bumps, but the invalidation key ("did anything change the outputs?") is not machine-checkable — a manually bumped RESULTS_VERSION risks silently stale results, and each request still pays report rendering + boundaries. **Deferred** unless /data traffic justifies it. |
[...]

Issue #293, opened 2026-04-14
API takes too long to create a community report from when user clicks


ejanalysis:
In the EJAM web app you can click to get a single-site report (from the map popup, or from the details tab site by site table, or from the excel  download version of the site by site table).

The API takes a while to respond, especially for a report on FIPS unit like a county - this should be improved. 

Currently the API has to start an R session then load the EJAM package and datasets and then run the analysis and create the report. Probably the delay is in the first couple of steps. The actual analysis and report can be slow for a very large analysis, especially for FIPS codes since they have to download boundaries first. But it can take too long to get that kind of report even for one county. 

Steps to recreate:
Navigate to: https://ejam2.policyinnovation.info/
Select "Select a category of locations" radio button
Choose "by Census place name ..." from drop down
Select Counties radio button
Enter/choose "Sheridan County, WY"
Press Done button
Start Analysis, progress elements display in turn
EJScreen Multisite Summary displays, with Community Report tab active, table data populated fine
Select Details tab
Clicking on the EJAM report in Column 2 opens a new tab that is blank for a very, very long time but eventually produces the EJAM multisite report

ericnost:
Yes, this is really an https://github.com/Public-Environmental-Data-Partners/EJAM-API issue. I recently upgraded the server config, increasing CPU and RAM, but it's not necessarily better...

ejanalysis:
Maybe paying for Posit Connect is worthwhile. They designed that product to be optimized for hosting very responsive shiny apps.

ejanalysis:
Last ranked: 2026-05-30
Cost score: 5
Benefit score: 19
Quadrant: B
Run report: inst/issues/ejam_issues_scored_by_risk_and_value.MD

ejanalysis:
Triage note: this looks very high value because community-report latency is a user-visible bottleneck in a core EJAM workflow.

ericnost:
Yea, the issue has always been:
1. finding dev ops expertise to recommend an optimal solution. I remember looking into posit connect a while back and thinking it wouldn't quite work, because we'd need to buy/host our own connect server. Didn't seem much different than trying to set up our own ArcGIS enterprise server. My sense is that there will always be some awkwardness is using EJAM to provide the community reports, compared to the previous method of a bespoke ArcGIS server.
[...]

PR #473, opened 2026-07-21
Trim the hardcoded PDF render waits, and make them settable without a rebuild

ejanalysis:
note: PR authored by Claude

Follow-up to the API latency work in Public-Environmental-Data-Partners/EJAM#293. PDF reports are roughly twice as slow as HTML, and a large part of that turned out to be fixed sleeping rather than work.

Draft: the measurements below are all from a developer laptop, and the one thing a laptop **cannot** tell us is whether the trimmed values are safe on the Cloud Run container. See "What could still go wrong".

## What was happening

Two pauses, neither of them a timeout — both are unconditional sleeps paid on **every** PDF whether or not they were needed:

- `webshot2::webshot(delay = 4)` when snapshotting the report map to PNG
- `pagedown::chrome_print(wait = 5)` when converting the report to PDF — `pagedown:::print_page()` literally calls `Sys.sleep(wait)`

That is ~9 seconds of guaranteed sleeping per PDF report.

They are not pointless, though. Each underlying tool waits for the page **load** event, but content that arrives *after* load — leaflet basemap tiles fetched over the network, JS-drawn content — is not covered by that. The pauses are padding for exactly that gap.

## Measurements

Against a real Kent County, DE report:

| what | result |
|---|---|
| report map snapshot, `delay` 0.2 → 4s | **pixel-identical at every value**, for both map shapes the report can produce (`mapfastej` and `ejam2map`) |
| PDF, `chrome_print` `wait` 0 → 5s | **identical at every value** — same 7 pages, same 23,027 chars, same 1,062 KB |
| end-to-end PDF, old (4/5) vs new (1/2) | **16.36s → 8.89s**, a 7.5s / 46% saving |
| all 6 pages of the old vs new PDF | **pixel-identical** (meanAbsPixelDiff = 0.000000 on every page) |

I also confirmed the map PNG produced at `delay = 1` is fully rendered — complete OSM basemap, county polygon, legend, attribution — rather than silently blank, and traced a real `ejam2report()` PDF run to confirm both new values are actually the ones being used (`delay=1`, `wait=2`).

## What changed

Defaults trimmed to **1s** (map snapshot) and **2s** (print) — deliberately *not* to the measured minimums of 0.2 and 0:

- **2s is pagedown's own default**, so that one returns to an upstream-chosen value rather than a number we invented.
- **1s keeps a 5× margin** over the 0.2s the map actually needed, because tile loading is network-dependent and a laptop with a warm tile cache is not the container.

New internal `pdf_wait_seconds()` reads both values from an R option or environment variable, so **a too-short default can be raised, or this change rolled back, on the server without rebuilding and redeploying the image**:

- `options(EJAM.pdf_map_snapshot_delay = 2)` or `EJAM_PDF_MAP_SNAPSHOT_DELAY=2`
- `options(EJAM.pdf_print_wait = 4)` or `EJAM_PDF_PRINT_WAIT=4`

The first usable value wins: **R option → environment variable → named package default in `global_defaults_package.R` → internal safety fallback**. Per review feedback, the defaults live as named entries in `global_defaults_package.R` rather than being set unconditionally as R options there, which would have shadowed the env-var rollback lever. Anything missing, non-numeric, non-finite, or negative is skipped rather than becoming a 0-second wait. Applied at all three call sites (two in `ejam2report()`, one in the app's PDF download handler) so the app and API stay in step.

## What could still go wrong

The failure mode here is **silent**: too short a pause still writes a valid PNG and a valid PDF, just with a blank basemap or missing content, and nothing errors. There is no automated test that would catch that, and adding a meaningful one would mean pixel-diffing rendered PDFs in CI.

Specifically not proven by these measurements:

1. **The container.** Slower CPU and a colder network than a laptop — the exact conditions where a short pause bites. This is why the values are settable at runtime.
2. **Large multisite reports.** I tested a single county (6–7 pages). A report with many sites has bigger tables and plots.

Given that, the safe rollout is to merge this and watch, with the env vars as the rollback lever, rather than treating the laptop numbers as proof.

## Note on sequencing

Nothing here reaches api.ejanalysis.com until an EJAM release is cut, the EJAM-API pinned version is bumped, and the image is rebuilt and redeployed. A ~7s saving does not justify that gate on its own — this should ride along with other API-affecting work (e.g. Public-Environmental-Data-Partners/EJAM#472 for Public-Environmental-Data-Partners/EJAM#446 layer 1) rather than triggering a deploy by itself.

Refs Public-Environmental-Data-Partners/EJAM#293


ejanalysis:
the option EJAM.pdf_map_snapshot_delay should get set in global_defaults_package.R ? but That should be addressed carefully: unconditionally setting the R option there would override the environment-variable rollback mechanism. try adding named package defaults there, with precedence:
R option → environment variable → global package default → safety fallback

ejanalysis:
Addressed in `5c708193`. Added named `pdf_map_snapshot_delay` and `pdf_print_wait` defaults to `global_defaults_package.R`, while preserving the first-valid-value precedence: R option → environment variable → package default → internal safety fallback. Invalid higher-priority values are skipped. Focused `ejam2report` tests: 63 passed, 0 failed, 0 warnings, 1 expected Pandoc skip.

— Codex

ejanalysis:
Independently verified the `5c708193` fix on a fresh worktree: precedence is R option → environment variable → named package default in `global_defaults_package.R` → internal safety fallback, exactly as requested. Since `.onAttach()` sources `global_defaults_package.R` into the session, the package-default layer resolves in app, console, and test contexts alike (via `global_or_param()`, which the helper wraps in `tryCatch`), and the R option is never set unconditionally, so the env-var rollback lever stays effective.

One small hardening in `c8251210`: a length-1 list set as the R option (e.g. `options(EJAM.pdf_print_wait = list("junk"))`) would have made `as.numeric()` error on the option path — which has no `tryCatch` — instead of being skipped like other unusable values. Now guarded with `is.atomic()` and covered by a test. Also aligned the `=` of the two new entries in `global_defaults_package.R`, and updated the PR description to state the full precedence chain.

Full `test-ejam2report.R` run after the change: 64 passed, 0 failed, 1 expected pandoc skip.

— Claude Code

PR #478, opened 2026-07-22
switch AOI

### LINKED ISSUE #477: switch AOI to ericnost/AOI
https://github.com/Public-Environmental-Data-Partners/EJAM/blob/ea476c8658e1b35caa6c4e69981503051953385b/DESCRIPTION#L121

Currently, mikejohnson51/AOI is a dependency install for EJAM. The issue is that AOI breaks for modern versions of R due to a namespace error. See: https://github.com/mikejohnson51/AOI/issues/29 This can, in turn, break an install of EJAM (in my experience). 

I'm not sure anyone will address the AOI issue directly, but I've created a patched fork.

ericnost:
Fixes #477

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -118,7 +118,7 @@ Suggests:
     zip
 Remotes:
     ejanalysis/ACSdownload,
-    mikejohnson51/AOI
+    ericnost/AOI
 VignetteBuilder: 
     knitr,
     rmarkdown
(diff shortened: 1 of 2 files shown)


ejanalysis:
This is very helpful. I am going to merge it into the development branch now too not just main. Normally I would merge to dev first instead of  directly into main, per the process I've been using with the EJAM repository. But this is a small fix that can go directly to main and dev at the same time.

ejanalysis:
closes #477

Issue #403, opened 2026-06-11
Add explanation of what "average in state" or "percentile in state" means for a multisite report


ejanalysis:
For a single site report it is obvious but for a multisite report the average in state and ratio to state average and percentile in state means something complicated. The doaggregate() function and others have to be able to present a ratio to state avg for a multisite report but there are multiple states, so code creates an average that makes sense. 

1) Consider how to explain this, such as using language like this (which was being used in the rarely-used function related to creating gt format tables, `table_gt_from_ejamit()` ): 

_Avg. in state means the average indicator value, among all the residents at these sites, using the statewide value in each resident's state. Percentile in state means the same, but using the site-specific value (expressed as a percentile) where each resident lives._

2) Consider where to put this info, when such ratios or percentiles are used for multiple site summaries: if it should be in a vignette, in webpages of documentation, as footnote on the summary report (UPDATE- That is done in v3.2022.2), on webpage(s) of results, in excel download, on plots, etc.

ejanalysis:
partly addressed via footnote in v3.2022.2 via PR #444 work

ejanalysis:
Reopening — this was closed automatically and unintentionally.

When the v3.2022.2 release merged to `main` (#495), commit 46ed4f04 from #444 carried `closes #403` in its commit message. Closing keywords in commit messages fire on their own once the commits reach the default branch, independently of the pull request body — and #495's body had deliberately listed this issue under *"Also addresses (still open by design)"*. The commit keyword won over the stated intent.

What actually shipped in v3.2022.2 is the summary-report footnote, which `NEWS.md` describes as *"part of #403 and #410"* — and which item (2) of this issue already records as `UPDATE- That is done in v3.2022.2`.

Still outstanding, from item (2)'s own list of venues:

- a vignette
- documentation webpages
- webpage(s) of results
- the Excel download
- plots

Item (1) — settling on the explanatory wording, e.g. the `table_gt_from_ejamit()` language quoted in the description — is only partly addressed by the footnote's phrasing.

No code change is needed to reopen; nothing in the release is affected.

Issue #391, opened 2026-06-03
Census API key now required (tidycensus >= 1.8 breaking change) — code + docs


ejanalysis:
## Background

**tidycensus 1.8** (breaking change): the Census Bureau now requires an API key for **all** requests, **including metadata endpoints** used by `load_variables()` and `get_pop_groups()`. tidycensus now **errors** (no longer warns) when no key is available. `load_variables()` / `get_pop_groups()` gained a `key` argument; `table=` lookups in `get_acs()` / `get_decennial()` forward the resolved key automatically.

For reference: tidycensus **1.7.0** (2025-01-23) set the default `get_acs()`/`get_pums()` year to **2023**. No release yet bumps the default to 2024 or adds 2024 ACS5 support — as of 1.8, `formals(tidycensus::get_acs)$year == 2023`.

## Impact on EJAM (tiered)

- **Released apps / end users — none.** ACS data is baked into package `.rda`/arrow; the running app never calls tidycensus, so the tidycensus version is irrelevant at analysis runtime.
- **Data-build pipeline — was at risk, now handled.** Bulk ACS data comes from `ACSdownload::get_acs_new()` (summary files), *not* tidycensus. The only tidycensus touch is `acs_table_info()` → `load_variables(yr, "acs5")` for table geography. Under 1.8 this *errors* without a key (previously warned), and the prior-year fallback (`all(is.na(geography))`) did not catch an error.
- **Auxiliary tidycensus-direct functions** (`acs_bybg()`, `acs_bycounty()`, `shapes_from_fips()`) — superseded by the `ACSdownload` path and/or have fallbacks; affected only for `year = 2024` while tidycensus's default year lags at 2023.

## Exposure points (audit)

| File | Census call | Old guard |
|---|---|---|
| `R/calc_blockgroupstats_acs.R` (`acs_table_info`) | `load_variables()` | **none** |
| `R/acs_bycounty.R`, `R/acs_bybg.R` | `get_acs` | warn "may require" |
| `R/shapes_from_fips.R` (×2) | `get_acs`, `tigris::places` | warn "may require" |
| `R/calc_bgwts_nationwide.R` | `get_decennial` | has `key` param |
| `R/download_bg_islandareas_raw.R` | passes `key = Sys.getenv("CENSUS_API_KEY")` | good pattern |

EJAM does **not** call `get_pop_groups()`.

[...]

PR #495, opened 2026-07-24
Release v3.2022.2: merge development into main

### LINKED ISSUE #127: web app details tab with table of all sites is too sluggish as it appears after clicking to view that tab
- is it the url-generating functions?

### LINKED ISSUE #156: in app - add view of   ejam2areafeatures(out)  (info on % of pop with certain things nearby)
only relevant to multisite summaries not 1-site report

This info is not easy to see anywhere else yet !!

and add info on this  to basics.Rmd   if not yet in app:  
```
out = testoutput_ejamit_100pts_1miles
ejam2areafeatures(out)   
```

### LINKED ISSUE #242: Report should show ratios to avg for special indicators like
These special indicator types also should get ratios to avg in community report (maybe not the ones that are counts/numbers of nearby features) -- see `ejam2areafeatures(testoutput_ejamit_100pts_1miles)` 
```
> names_flag
[1] "yesno_tribal"    "yesno_impwaters" "yesno_airnonatt" "yesno_cejstdis"  "yesno_iradis"   
> names_featuresinarea
[1] "num_school"   "num_church"   "num_hospital"
> names_sitesinarea
[1] "count.NPL"      "count.TSDF"     "num_airpoll"    "num_tri"        "num_waterdis"   "num_brownfield"
```

Consider how/where to show ratios for these -- _count_ of schools doesn't need a ratio, but maybe _% of sites analyzed here_ that have a school nearby  vs % of US -- that is useful.  see `ejam2areafeatures(testoutput_ejamit_100pts_1miles)` 

```
CRITICAL SERVICES
Flag for Overlapping with Housing Burden Communities	1	NA	NA
Flag for Overlapping with Food Desert Areas	1	NA	NA

FEATURES AND LOCATION INFORMATION
Number of Sites Nearby (total unique)	1	NA	NA
Distance to Closest Site for Avg Person	2	NA	NA
Number of Sites Nearby (avg)	1	NA	NA
Number of Sites Nearby (max)	1	NA	NA
Distance to Closest Site	0	NA	NA
Count of National Priority List Superfund sites	0	NA	NA
Count of Treatment Storage Disposal Facilities (TSDF)	6	NA	NA
Number of Air Pollution Facilities	78	NA	NA
Number of Toxic Release Facilities	5	NA	NA
Number of Water Discharge Facilities	45	NA	NA
Number of Brownfields	0	NA	NA
Number of Schools	15	NA	NA
Number of Worship Places	28	NA	NA
Number of Hospitals	4	NA	NA
Flag for Overlapping with Tribes	0	NA	NA
Flag for Overlapping with Impaired Waters	1	NA	NA
Flag for Overlapping with Non-Attainment Areas	1	NA	NA
Flag for Overlapping with CEJST Disadvantaged Communities	1	NA	NA
Flag for Overlapping with EPA IRA Disadvantaged Communities	1	NA	NA
```

see #241 but these are different/special case

### LINKED ISSUE #348: API report links from multisite results label every per-site report as site 1
Per-site "EJAM Report" links generated for `results_bysite` omit the original site number. The current `GET /report` endpoint then regenerates a one-site analysis and calls `ejam2report(..., sitenumber = 1)`, so reports opened from site 2, site 3, etc. can be labeled as site 1.

Expected behavior:
- Per-site report links preserve the original row/site label, or omit the misleading site number and show a stable site id/name instead.
- A per-site API report opened from row N should not imply it is row/site 1 unless it really is.
- Add coverage that a per-site API report link generated for row N either carries a display-site identifier or produces report header text that does not incorrectly say site 1 for every row.

Evidence from ACS2024 triage:
- `url_ejamapi()` omits `sitenumber` for per-site URLs.
- `inst/plumber/plumber.R` hard-codes `sitenumber = 1` for `GET /report`.

Related open issues: #22 is the broad API umbrella, #126 covers API-query reports for one-site/multisite polygon/shapefile workflows, and #149 is another `url_ejamapi()` sitenumber cleanup.

Triage note: This issue was added based on 5/2026 notes triage.

### LINKED ISSUE #385: ejam2report() fails to save a file when the filename parameter is used
This saves a report in a temp dir:
`
x = ejam2report(testoutput_ejamit_10pts_1miles , launch_browser = FALSE )
file.exists(x)
`
But this fails to save a report:
`
x = ejam2report(testoutput_ejamit_10pts_1miles , launch_browser = FALSE, filename = "junk.html" )
file.exists(x)
`

### LINKED ISSUE #410: "FEATURES AND LOCATION INFORMATION" on report need explanation and less rounding
- [x] Fix Rounding (of feature-related indicators in report): If we show counts as average person's count, as we have, we need to round to 1 or two decimals not zero decimals, since rounding to zero decimals would show value of 0.4 as 0, even though 0.4 means 40% of the residents have the given feature within their blockgroup. If we start to show counts as totals, not averages, then rounding to zero decimals is OK. rounding is controlled by map_headernames$decimals as explained in the comment below on this issue.

- [x] Edit Text Labels and Footnote (of feature-related indicators in report): Improve the text labels or header or explanation of the community report rows like "Count of National Priority List Superfund sites" and "Flag for Overlapping with CEJST Disadvantaged Communities" -- Change them to clarify this: 
  - [x] 1) make clear that the counts are the average resident's count, and --NOTE THIS WAS INCORRECT ACTUALLY
  - [x] 2) make clear that the counts are counts of the feature anywhere within the resident's blockgroup, not counts anywhere at these sites, and not counts within the resident's blockgroup limited to only the part of the blockgroup that is at these sites.  This clarification probably requires a footnote under the table to explain clearly, and also an improved text label on each line. For example, it would be more clear but repetitive to say, "Number of Schools (in Avg. Resident's Blockgroup)." Also, "Flag for Overlapping with ..." really means "

(also see #242 which is about showing most of these indicators in terms of % of people with one or more schools, etc. within their blockgroup).

The relevant indicators might include these:
```
vars <- c(names_sitesinarea, names_featuresinarea,  names_flag,      names_criticalservice)
cbind(varlist = varinfo(vars)$varlist, rname = vars, longname = fixcolnames(vars,'r','long'))
```
The relevant part of the report looked like this as of v3.2022.0 June 2026:
Image

### LINKED ISSUE #465: EJAM app crashes on EJScreen token handoff (Send to EJAM) for FIPS/polygon/bufferless selections
## Summary

The EJAM Shiny app **crashes** (session dies → "Sorry ... This app has stopped because of an error or a timeout") when the EJScreen Report tool hands off a selection via **"Send to EJAM"** (the token-based `?handoff=<token>` path) for any selection that carries **no buffer/radius** — i.e. every **County/Tract (FIPS)** and **drawn-polygon** selection, and lat/lon selections made without a buffer.

The **"Multisite Report"** button on the same EJScreen selection works fine, because it goes straight to the EJAM API (`GET/POST /report`) and never launches the Shiny app.

Reported by @ejanalysis.

## Steps to reproduce

1. Open https://pedp-ejscreen.azurewebsites.net/index.html → **Reports** → **Select an Area** → **Select County** → **+ Add to Multisite list**.
2. Click **Send to EJAM**.
3. The EJAM app opens at `…/?handoff=<token>` and immediately shows the crash overlay ("This app has stopped because of an error or a timeout" / "Click to Restart the App").

Also reproduces for **Select Tract** and **draw polygon**. Reproduces directly by minting a token against the production API and opening the app:

```
# POST returns a token; opening the app with it crashes the session
curl -s -X POST https://api.ejanalysis.com/handoff \
  -H "Content-Type: application/json" -d '{"method":"FIPS","fips":["10001"]}'
# -> https://ejam.publicenvirodata.org/?handoff=<token>   (crashes)
```

Confirmed live against production EJAM **v3.2022.1**.

## Root cause

The launch-URL handoff observer in `R/app_server.R` fetches the payload from the EJAM API's `GET /handoff/<token>` and parses it with `jsonlite::fromJSON()`.

The API (`Public-Environmental-Data-Partners/EJAM-API` `rest_controller.r`) stores the payload as `list(method, sites, fips, shape, radius)` and returns it verbatim. Any field that was **absent** in the incoming selection is an R `NULL`, and the plumber JSON serializer emits `NULL` as **`{}`**. So a FIPS handoff comes back as:

```json
{"method":["FIPS"],"sites":{},"fips":["10001"],"shape":{},"radius":{}}
```

`jsonlite::fromJSON()` parses each `{}` as a **zero-length list**, not `NULL`. The observer's radius block then does:

```r
if (!is.null(spec$radius)) {
  radval <- suppressWarnings(as.numeric(spec$radius))   # as.numeric(list()) -> numeric(0)
  if (!is.na(radval) && radval >= 0) url_radius(radval)  # if() on a zero-length value -> ERROR
}
```

`spec$radius` is a non-`NULL` empty list, so the guard passes; `as.numeric(list())` is `numeric(0)`; and `if (!is.na(numeric(0)) && …)` evaluates `if()` on a zero-length condition → **"missing value where TRUE/FALSE needed"**. This is an **unhandled error in a `priority=1000` observer during session init**, which terminates the Shiny session — the crash.

**Why lat/lon "worked":** EJScreen's `multisite.js` only puts `radius` in the handoff payload when a buffer is set (`if (r) payload.radius = r`). FIPS and polygon baskets never carry a buffer, so `radius` is always `{}`. The earlier lat/lon handoff verification happened to use a non-zero buffer, so `radius:[N]` parsed to a real scalar and dodged the bug. A lat/lon handoff with no buffer crashes too.

**Why direct `?fips=` deep-links work:** the non-token branch sets `spec$radius <- q$radius %||% q$buffer`, which is a clean `NULL` when absent (not an empty list), so the same `if()` is skipped.

## Fix

In the EJAM app (this repo — no API/EJScreen redeploy needed, and it hardens the app against any malformed handoff payload):

1. Normalize the parsed handoff payload so zero-length (`{}`) fields become `NULL`.
2. Make the radius block length-safe (`length(radval) == 1 && …`) as defense-in-depth.

PR into `development` follows.

### LINKED ISSUE #467: ejam2report() fails for zero-population or otherwise analysis-invalid sites

ejanalysis:
## EJAM v3.2022.2 (August 2026)

Cuts the v3.2022.2 release: 52 commits of development since v3.2022.1. Full changelog in the top section of `NEWS.md`. This is a code-and-docs patch — it reuses the existing `ejamdata` **v3.2022.0** data release (no data change; `ejamdata_required_tag` stays v3.2022.0).

**Highlights**
- **New**: State versions of the flagged-areas stats, now shown in the community report, Excel "Area Features" tab, and barplots (#444)
- **~4x faster** Site-by-Site Table in the web app's Details tab (#491), ~46% faster PDF reports (#473), county boundaries baked into the package (#472)
- **Fixes**: EJScreen "Send to EJAM" handoff crash (#466), empty saved reports when `filename=` given (#471/#475), zero-population sites (#468), per-site links all labeled "Site 1" (#470/#479), percent-display metadata repairs (#488), `acs_bybg(year=2024)` (#469)
- **Internal**: all key URLs single-sourced in DESCRIPTION via `url_package()` (#485); AOI dependency now `ericnost/AOI` (#478); "Draft API Endpoints" Swagger grouping (#509); regenerated example datasets and a new release workflow (#512)

## Closes (restated here because keywords only fire on main)

Closes #127
Closes #242
Closes #348
Closes #385
Closes #465
Closes #467
Closes #156
Closes #410

Also addresses (still open by design): #403, #446 (layer 1, county bounds is done here), #293 (latency work). #477 and #391 were already closed.

**Deferred to v4+ (not in this release):** #483 zipcode analysis; API-in-EJAM; htmltemplate_report.

## Pre-merge checklist

- [x] `devtools::check()` passes — all unit tests + all Rd examples (**0 ERRORS / 0 WARNINGS / 4 benign NOTES**)
- [x] Vignette examples tried where feasible (all 26 vignettes rebuilt in check)
- [x] Failures/warnings/notes reported and triaged
- [x] pkgdown site / vignettes proofread
- [x] Version fields verified consistent (DESCRIPTION 3.2022.2 / August 2026; CITATION.cff; `_pkgdown.yml`; golem-config; NEWS heading dated; ejamdata pin v3.2022.0)
- [x] **Full testoutput regeneration done** (#512) — see below

## The testoutput regeneration gate — now satisfied

This PR originally would have shipped the release's headline feature looking broken. All 12 shipped `testoutput_*` datasets predated #444, so their `flagged_areas` tables had **no** `Percent_of_all_People_Statewide` and no `ratio_to_state_avg` column, and every example, vignette, and stored example report displayed **N/A** for the new State flagged-areas stats.

Fixed in #512, now merged into development:

- Re-ran the points (10/100/1000 pts @ 1 mi), fips (counties + cities), and shapes families, keeping the existing test points so results stay comparable. Regenerated the matching `.xlsx` / `.html` / `.pdf` example outputs.
- **Values were spot-checked, not just column presence.** The 100-point and shapefile analyses show properly varied State ratios. `testoutput_ejamit_fips_counties` reads exactly `1.00` on every row, which is correct rather than a self-referential bug: its input is FIPS 10001/10003/10005, all three Delaware counties, so the analyzed area *is* the whole state and the analyzed percentage equals the State percentage by construction.

| | before regeneration | after |
|---|---|---|
| `devtools::check()` | 0 ERROR / 0 WARN / 4 benign NOTE | identical |
| tests | pass (~431s) | pass, no failures |
| `devtools::build()` | — | `EJAM_3.2022.2.tar.gz`, 188.6 MB |

## Also folded in since this PR was opened

- **#509** — "Draft API Endpoints" Swagger grouping for the bundled local API. Merged to development independently; it ships in this release, so it is now recorded in `NEWS.md`.
- **#512** — regenerated datasets, a halved README "What's New" section, and `.github/workflows/release.yaml`.

## After merge

1. Tag **`v3.2022.2`** on main and publish the GitHub Release (a draft is prepared for review).
2. Merge #493 (→ dev-deploy; deploys dev app) → verify at the dev app
3. Un-draft + merge #494 (→ prod-deploy; deploys production)
4. EJAM-API: bump its `EJAM_VERSION` build-arg when redeploying

> **Note on the release workflow:** scheduled workflows only fire from the default branch, so `release.yaml` becomes active for the August 1 run only once this PR merges. It skips automatically if a tag *or* a release (including an unpublished draft) already exists for the version, so a hand-reviewed draft release will not be duplicated or clobbered.


ejanalysis:
## Pre-merge gate: check() triage (per the release checklist + dev-update-package.Rmd)

Ran full `devtools::check()` on the candidate (with vignettes; pandoc + Census key in env). **First result: 3 ERRORs / 0 WARNINGs / 5 NOTEs — after triage, zero were release regressions**, and the package-side items are fixed in `7a7d564f`:

| Finding | Verdict | Action |
|---|---|---|
| examples ERROR: `url_acs_table_info()` needs a Census API key | Pre-existing (same on main); fails only for keyless users | Wrapped in `\donttest` with a key note; verified it runs with a key (16 URLs) |
| `--run-donttest` ERROR: `acs_bycounty()` needs a key | Environmental (key absent in the stripped check env) | Verified with key: 62 NY county rows. No change needed |
| tests ERROR: `test-shiny-1-14-compat.R` reads `../../R/app_ui.R` | Real, but dates to #416 — source-inspection test can't run against the installed package | Now skips visibly under R CMD check; DESCRIPTION shiny-floor check falls back to installed DESCRIPTION |
| spelling snapshot diff | New legit jargon from this cycle's docs | `spelling::update_wordlist()` (+85/−26) |
| NOTE: stray top-level `figure/` | 4 committed knitr PNGs | `.Rbuildignore`d (deleting them is a separate call) |
| NOTEs: `:::` self-calls, NSE no-visible-binding, 2 long Rd lines, `../reference/` links in vignettes | Benign/pre-existing | Left as-is |

**metadata_check()** run per dev-update-package.Rmd: expected mix for a code-only patch (untouched data stamped 3.2022.0 by design). Only oddity: `ejampackages` (trivial internal helper) stamped acs 2020-2024 — noted, not a blocker.

**pkgdown proofread** (rendered /dev/ site): changelog shows one clean dated "EJAM 3.2022.2 (August 2026)" section (no stray "development version" heading); dev-api article's executed example outputs `https://api.ejanalysis.com/report?fips=3650617,0973000&buffer=0&sitenumber=0&fileextension=html`; footer reads "August 2026 / EJAM Version 3.2022.2"; webapp article clean.

**Confirmation `check()` re-run is in progress** on the fixed candidate (with key + pandoc); result will be posted here. CI on this PR so far: pkgdown ✅, quick-install ×2 ✅, test-webapp-functionality ✅.

ejanalysis:
## ✅ Pre-merge gate COMPLETE — final `devtools::check()` is green

**ERRORS: 0 | WARNINGS: 0 | NOTES: 4 (benign)** — testthat: **FAIL 0 | PASS 4681 | SKIP 28** (skips are the deliberate source-tree/CRAN guards). Full run with vignette rebuilds (all 26), Rd examples incl. `--run-donttest` (with a Census key), and the whole test suite.

Second-round fixes that got here (`a63139a6`, plus `6e51b941` deleting the stray `figure/` PNGs):

- **Regenerated `testoutput_doaggregate_{10,100,1000}pts_1miles`** — their `longnames` predated #444/#488's deliberate indicator renames. Regeneration was gated: verified exactly 4 longnames changed and **zero data values differ** per fixture before overwriting (per the release checklist in `dev-update-package.Rmd`).
- **3 bare `desc::desc_get("VersionACS")` calls** in tests anchored to the installed DESCRIPTION (they errored under R CMD check's working dir); recipe-runner test now skips visibly when `.Rbuildignore`d `data-raw/` is absent.
- **Geocoder expectations updated** — the live ArcGIS geocoder now returns ZIP 20004 (correct street ZIP for 1200 Pennsylvania Ave NW) vs the saved 20044 (old PO-box ZIP); 7 strings updated + `tolerance = 0.01` for match-score drift. All 7 pass live.

Remaining 4 NOTES (unchanged, benign, pre-existing): `:::` self-calls (2 objects), NSE no-visible-binding in pipeline report fns, 2 long Rd lines (PDF manual only), pkgdown-style `../reference/` links in vignettes.

**Every box in the pre-merge checklist is now ticked.** From here (per the release plan): merge this PR → tag `v3.2022.2` on main → GitHub release using NEWS.md as changelog → #493 (dev deploy) → verify → #494 (prod).

ejanalysis:
## ✅ Gate re-verified after the 2026-07-24 merges — still green

Development gained 9 commits since the original gate run (#499 local-API mirror, #501 docs URLs, #502 url_package errors, #503 Config/EJAM field migration, #505 ?advanced= param, ratio-columns-by-default, post-merge review fixes, NEWS), so the full `devtools::check()` was re-run on the current candidate (`3d9d4617`):

**ERRORS: 0 | WARNINGS: 0 | NOTES: 4 (same benign set)** — testthat: **FAIL 0 | PASS 4712 | SKIP 29** (test count up from 4681 with today's new tests; skips are the deliberate source-tree/CRAN/opt-in-server guards).

Also today: all review threads on the merged PRs resolved (including 3 post-merge Copilot findings, fixed in `54f8905e`); the RC (`EJAM@development`) is deployed on the dev server — full app (isPublic=FALSE), ratio columns on by default, PDF Chrome fix — awaiting Mark's final pass before this PR merges.

ejanalysis:
## ✅ Release gate re-verified on the final candidate (`23d83235`) — green

Re-ran the full `devtools::check()` after today's report fixes landed:

**ERRORS: 0 | WARNINGS: 0 | NOTES: 4** (the same benign set) — testthat **FAIL 0 | PASS 4712 | SKIP 29**.

Additional pre-release double-checks, all passing:

| Check | Result |
|---|---|
| Version fields (DESCRIPTION, CITATION.cff, _pkgdown.yml, golem-config, NEWS heading, ejamdata pin + marker) | all consistent: **3.2022.2 / August 2026 / 2026-08-01**, ejamdata **v3.2022.0** |
| `metadata_check()` (per `dev-update-package.Rmd`) | expected mixed-vintage stamps for a code-only patch |
| Map popup escaping fix | verified **end-to-end** in a rendered FIPS report: zero `&lt;a href` in the widget; injected `<script>` in another column still escaped (XSS protection intact) |
| PDF page breaks | verified in a **rendered PDF** (pdftools page-text analysis): Limited-English (p4), flagged-areas (p5), Facility Counts (p6) each begin at the top of a page |
| Report suites | build_community_report 5/5, ejam2report 29/29, utils_flagged_FUNCTIONS 19/19, MAP_FUNCTIONS 36/36 |

The dev app is being re-deployed from `development` so the final review round includes these fixes. **Still not merging this PR / not tagging** — release remains scheduled for Aug 1 per plan.

ejanalysis:
## Pre-merge gate status — one item still open, and it blocks this PR

**The full testoutput regeneration gate was not satisfied when this PR was opened.** I verified it directly: all 12 shipped `testoutput_*` datasets still had `flagged_areas` tables with no `Percent_of_all_People_Statewide` and no `ratio_to_state_avg` column. Merging as-is would have shipped **N/A** for the State flagged-areas stats in every example, vignette, and stored example report — the headline feature of this release.

That work is now done and verified, but it lives in **#512, which is not yet merged into `development`.** Until #512 lands, this PR does *not* contain the fix.

### What #512 contains

- Regenerated points (10/100/1000 pts @ 1 mi), fips (counties + cities), and shapes families, keeping the existing test points so results stay comparable, plus the matching `.xlsx` / `.html` / `.pdf` example outputs. All four `ejamit` families now carry both new columns, and the values were spot-checked rather than just the column names.
- A short README "What's New in v3.2022.2" section (README had no per-version section at all).
- `.github/workflows/release.yaml`, which tags and publishes the release from `DESCRIPTION` + `NEWS.md`, either on demand or on a 13:00 UTC August 1 schedule.

### Verification

| | before regeneration | after |
|---|---|---|
| `devtools::check()` | 0 ERROR / 0 WARN / 4 benign NOTE | identical |
| tests | pass (~431s) | pass, no failures |
| `devtools::build()` | — | `EJAM_3.2022.2.tar.gz`, 188.6 MB |

### Two things to know before merging this PR

1. **#512 must merge first**, or the regenerated data and the release workflow are not in this release. A plain squash merge of #512 is currently refused — GitHub reports "the base branch policy prohibits the merge" — so it needs either `--admin` or auto-merge. I deliberately did not bypass the ruleset without a decision from a human.
2. **The August 1 automation only works if `release.yaml` reaches `main`**, because scheduled workflows fire only from the default branch. That means it has to ride along in this PR.

PR #502, opened 2026-07-24
Improve url_package invalid type errors

ejanalysis:
## What changed

- Replaced the assertion-style `stopifnot()` check for `url_package(type = ...)` with an explicit, user-facing error.
- The error reports the supplied invalid value, lists every accepted `type`, and points users to `?url_package`.
- Added focused assertions for the invalid value, the beginning and end of the valid-option list, the help pointer, and multi-value input.

## Why

The previous assertion message exposed the implementation condition but did not tell users which values were valid or where to find the function documentation.

## User impact

Invalid calls now produce guidance such as:

```
Invalid `type` value: "not-a-real-type". `type` must be one of: "code", ... "api". See ?url_package.
```

## Validation

- Sourced `R/url_package.R` and ran `testthat::test_file("tests/testthat/test-url_package.R")` — 65 passed, 0 failed/warned/skipped
- Manually verified the complete invalid-type error text
- `git diff --check`

The broader `devtools::test(filter = "url_package")` entry point was blocked during package startup because the isolated worktree does not contain EJAM's ignored local Arrow datasets and its fallback dataset download failed; the focused test file itself passed completely using the sourced implementation.

— Codex

### DIFF
--- a/R/url_package.R
+++ b/R/url_package.R
@@ -213,7 +213,21 @@ url_package <- function(
   # (get_full_url = FALSE) makes sense
   repo_types <- c("code", "ejamrepo", "ejscreenrepo", "apirepo", "data", "datarepo")
 
-  stopifnot(length(type) == 1, type %in% names(field_by_type))
+  valid_types <- names(field_by_type)
+  if (length(type) != 1 || is.na(type) || !(type %in% valid_types)) {
+    supplied_type <- if (length(type) == 0) {
+      "<empty>"
+    } else {
+      paste(encodeString(as.character(type), quote = "\""), collapse = ", ")
+    }
+    stop(
+      "Invalid `type` value: ", supplied_type, ". ",
+      "`type` must be one of: ",
+      paste(encodeString(valid_types, quote = "\""), collapse = ", "),
+      ". See ?url_package.",
+      call. = FALSE
+    )
+  }
   stopifnot(length(desc_or_alias) == 1, desc_or_alias %in% c("desc", "alias", "redirect"))
 
   if (desc_or_alias %in% c("alias", "redirect") && get_full_url == FALSE) {
--- a/R/url_package.R
+++ b/R/url_package.R
@@ -182,7 +182,10 @@ url_package <- function(
     domain = NULL
 ) {
 
-  if (all(type %in% c("github.com", "github.io"))) {
+  # length guard: all(character(0) %in% ...) is vacuously TRUE, so without it an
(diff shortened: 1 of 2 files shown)


R/url_package.R
```diff
@@ -213,7 +213,21 @@ url_package <- function(
   # (get_full_url = FALSE) makes sense
   repo_types <- c("code", "ejamrepo", "ejscreenrepo", "apirepo", "data", "datarepo")
 
-  stopifnot(length(type) == 1, type %in% names(field_by_type))
+  valid_types <- names(field_by_type)
+  if (length(type) != 1 || is.na(type) || !(type %in% valid_types)) {
```
ejanalysis, line 230: This validation occurs after the legacy-domain compatibility block, so `url_package(character())` never reaches the intended `<empty>` error. `all(character(0) %in% ...)` is `TRUE`, `domain <- type[1]` becomes `NA`, and the earlier `if (all(domain == "github.com"))` throws `missing value where TRUE/FALSE needed` (plus an obsolete-parameter warning). I reproduced that behavior on this head. Please move the cardinality/type validation ahead of the compatibility block, or guard that block with `length(type) == 1L`, and add `character()` to the focused tests.

— Codex

ejanalysis (commented):
Quick review found one invalid-input edge case in the new validation; I reproduced it on this exact head and left the actionable detail inline. Apart from that case, the error-message change is focused, and this head combines cleanly with Public-Environmental-Data-Partners/EJAM#503.

— Codex

R/url_package.R
ejanalysis: Fixed in f088f2ab: guarded the legacy block with `length(type) == 1L` (validation stays AFTER the block so legacy `type="github.com"` still remaps to `"code"` before being validated), and added `url_package(character())` to the focused tests — it now produces the intended `Invalid `type` value: <empty>` error. Verified by running the test file (0 failures).

ejanalysis (commented):
**Claude-fix follow-up review** — reviewed final head `08e2cb0f` and the resulting `development` integration.

The empty-input issue is fixed without breaking the legacy `github.com`/`github.io` compatibility path: the legacy block now has a scalar-length guard, and the focused tests cover `url_package(character())`. I independently reproduced the intended `Invalid `type` value: <empty>` error on the merged code. The fix is also preserved through the Config/EJAM migration. No new actionable issue found. GitHub reports no check runs on this PR.

— Codex

PR #533, opened 2026-07-26
Correct the web app settings vignette on main before the release

### LINKED ISSUE #109: Update & correct the article "Defaults and Custom Settings for the Web App"
This article/vignette was drafted before more changes were made in how it all works so some text is wrong now. eg, "radius" is the new name for the ejamapp() parameter not "radius_default." Fix that.

The examples in help doc on [ejamapp](https://ejanalysis.github.io/EJAM/reference/ejamapp.html)are useful and updated so could not repeat those (delete examples in vignette) & just point user to help on ejamapp() function.

Another way to make this more up-to-date is to change some R code blocks to say eval = TRUE and delete the pasted results snapshot that was shown as commented lines.

See article here:
https://ejanalysis.github.io/EJAM/articles/dev-app-settings.html

generated based on this file:

https://github.com/ejanalysis/EJAM/blob/HEAD/vignettes/dev-app-settings.Rmd

via EJAM:::pkgdown_update() or iterate via pkgdown::build_article()

ejanalysis:
Carries `7d8ddb30` (#526, merged to `development`) to `main` so the corrected vignette ships in v3.2022.2 rather than sitting a release behind.

`vignettes/dev-app-settings.Rmd` only — 212 insertions, 977 deletions. No code, data, version, or workflow change. After this, `main` and `development` differ only by the `inst/issues/` ranking artifacts, which stay on `development` by design.

**Verified without a local build:** the pkgdown workflow ran on `7d8ddb30` on `development` and succeeded. `pkgdown::build_articles()` executes the vignettes, including the three chunks here that override the file's global `eval = FALSE`, so this file is already known to render. That check matters because `check-standard` (R CMD check) is `disabled_manually`, so nothing else in CI would currently catch a broken vignette.

Safe with respect to the release: it does not touch `DESCRIPTION`, `NEWS.md`, the draft release, or `release.yaml`, so the August 1 scheduled publish is unaffected.

### DIFF
--- a/vignettes/dev-app-settings.Rmd
+++ b/vignettes/dev-app-settings.Rmd
@@ -1,7 +1,7 @@
 ---
 title: "Defaults and Custom Settings for the Web App"
-description: "Defaults and Custom Settings for the Web App"
-output: 
+description: "How EJAM app defaults, launch parameters, bookmarks, and Advanced Settings differ"
+output:
   rmarkdown::html_vignette:
     toc: true
 vignette: >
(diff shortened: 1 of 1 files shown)


ejanalysis:
closes #109

PR #540, opened 2026-08-03
Release workflow: start pkgdown and the install check itself

ejanalysis:
Fixes the gap found when v3.2022.2 shipped: **publishing a release from this workflow triggers nothing downstream.**

## What went wrong

GitHub does not create workflow runs from events raised by the default `GITHUB_TOKEN`. `release.yaml` publishes with that token, so the `release: published` event it generates never reached `pkgdown.yaml` or `install-release-user-check.yaml`.

Observed 2026-08-01: v3.2022.2 published cleanly at 21:44 UTC, and **neither workflow ran**. The docs root still served 3.2022.1 two days later, until both were dispatched by hand. Nothing was red — the release run itself succeeded, and the *absence* of two runs was the only symptom, which is what made it easy to miss.

## The fix

A final step dispatches both workflows. `workflow_dispatch` is the documented exception to the `GITHUB_TOKEN` rule, so this needs **no PAT and no new secret** — only `actions: write`, added to the permissions block.

## Two things worth reviewing

**It re-implements pkgdown's "Latest" gate instead of trusting it.** Passing `target_folder` explicitly overrides that gate on pkgdown's side ("a manual workflow_dispatch target always wins"). Without the check added here, publishing a non-recommended line — v3.2024.0 exists while v3.2022.x is what users are pointed at — would overwrite the root docs. So the gate is enforced before the dispatch, with the same 5-attempt retry pkgdown uses, since `/releases/latest` can lag publication. The install check is not gated; it runs for every release.

**A dispatch that fails to start fails the step**, with the manual commands written into the job summary. By then the release is published and a red run does no harm — whereas a silent miss is exactly the failure being fixed.

## Verification

- YAML parses; steps and permissions confirmed via `YAML.load_file`.
- The step's shell script passes `bash -n`.
- Dry runs are unaffected: the step carries the same `!(workflow_dispatch && dry_run)` guard as the publish steps, so `dry_run=true` still touches nothing.
- No local machine paths in the diff.

Cannot be end-to-end tested without publishing a release. The next release is the real test; if it fails, the job summary says exactly what to run by hand.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -146,8 +146,30 @@ jobs:
             fi
           fi
 
+          # Whether a published release for this version exists once this job is
+          # done - either because a step below publishes it, or because it was
+          # already published. This is what gates the downstream dispatches, and
+          # it is deliberately NOT the same as "skip".
+          #
+          # Using skip there would make a failed dispatch unrecoverable: the
+          # release is published by the time the dispatch runs, so a rerun sees
+          # already_published -> skip=true, quietly does nothing, and finishes
+          # green with the follow-up workflow still never started. That is the
+          # exact silent gap this workflow is meant to close. Rerunning is
+          # therefore the supported recovery path, and re-dispatching for an
+          # already-published tag is harmless: pkgdown rebuilds identical docs
+          # and the install check just runs again.
+          #
+          # A scheduled run that bails on a version mismatch keeps mode=create
+          # with SKIP=true, so it correctly dispatches nothing.
+          RELEASED=false
+          if [ "$MODE" = 'already_published' ] || [ "$SKIP" = 'false' ]; then
+            RELEASED=true
+          fi
+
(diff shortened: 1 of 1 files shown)


ejanalysis (commented):
### [P2] Keep failed downstream dispatches recoverable on rerun

At `.github/workflows/release.yaml:323-324`, if either `gh workflow run` call fails, this step exits nonzero after the release has already been published. On a standard rerun, the guard at lines 138-145 sees the release as `already_published`, sets `skip=true`, and this condition skips the dispatch step. The rerun can therefore finish green while the missing workflow was never started—the same silent gap this PR is intended to prevent.

Please make follow-up dispatches retryable for an already-published tag, or add an explicit follow-up-only recovery path. The failure summary should also print only the command(s) that actually failed, since it currently lists both and may duplicate the dispatch that succeeded.

The core `workflow_dispatch` approach and Latest-release gate otherwise look sound, and both current checks pass.

— Codex

PR #541, opened 2026-08-05
Sync release.yaml from main to development (kill the drift)

ejanalysis:
`release.yaml` was edited directly on `main` during the v3.2022.2 release and never carried back, so `development` has been holding a stale copy. Left alone, the next `development → main` release sync would have **reverted those edits** — losing fixes that were made precisely because the release exposed gaps.

Brings over #539 and #540, taking main's file verbatim:

- **Schedule moved** from 13:00 to 21:00 UTC (09:00 → 17:00 EDT).
- **The workflow now starts pkgdown and the install check itself.** This is the interesting one: GitHub does not create workflow runs from events raised by the default `GITHUB_TOKEN`, so the `release: published` event this workflow generates **triggers nothing on its own**. The original design assumed that event would drive the docs rebuild; it doesn't. Hence the explicit dispatch.
- **`RELEASED` gating** separate from `skip`, so a failed dispatch is recoverable by rerunning instead of silently short-circuiting on `already_published`.

No local edits — `git checkout origin/main -- .github/workflows/release.yaml`. YAML parses.

After this, `main` and `development` differ only by the `inst/issues/` ranking artifacts, which live on development by design.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -101,11 +101,16 @@ jobs:
           set -euo pipefail
           VERSION='${{ steps.ver.outputs.version }}'
           SKIP=false
+          # Tracked separately from SKIP because a scheduled version mismatch must
+          # suppress the downstream dispatches too, and SKIP alone cannot express
+          # that - see the RELEASED comment below.
+          SCHEDULE_VERSION_MISMATCH=false
 
           if [ '${{ github.event_name }}' = 'schedule' ]; then
             if [ "$VERSION" != "$SCHEDULED_EXPECTED_VERSION" ]; then
               echo "::notice::Scheduled run is a no-op. DESCRIPTION on main is $VERSION but this schedule only ships $SCHEDULED_EXPECTED_VERSION."
               SKIP=true
+              SCHEDULE_VERSION_MISMATCH=true
             fi
           else
             EXPECTED='${{ inputs.expected_version }}'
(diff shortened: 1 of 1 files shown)


ejanalysis:
Good catch — this was a real bug, and the code comment asserted the opposite of what the code did.

`MODE` is decided *after* the version-mismatch check, and purely from what is on the releases page. So on a future annual firing — `DESCRIPTION` long past the pinned version, that version's release already published — the run would print *"Scheduled run is a no-op"*, then land on `MODE=already_published`, set `RELEASED=true`, and dispatch pkgdown and the install check anyway. Not destructive, but it burns ~16 minutes of CI on a run that just announced it was doing nothing, and it contradicts the one-shot design.

Fixed in `fdc83b85`: a scheduled version mismatch is now tracked in its own flag (`SCHEDULE_VERSION_MISMATCH`) and excluded from the dispatch gate. The comment now describes the actual mechanism instead of claiming `mode` stays `create`.

Simulated every path that matters:

| event | version vs pin | release state | dispatch |
|---|---|---|---|
| schedule | **mismatch** | published | **false** ← was `true` |
| schedule | mismatch | none | false |
| schedule | match | draft | true |
| schedule | match | none | true |
| workflow_dispatch | match | **already published** | **true** |
| workflow_dispatch | match | none | true |

That second-to-last row is the one worth keeping an eye on: a manual rerun when the release is already published must still dispatch, because that is the recovery path for a dispatch that failed after publication — the exact silent gap `RELEASED` was split from `SKIP` to close. It still works.

**Follow-up needed:** this PR was meant to copy `main`'s file verbatim, so the same bug is live on `main` right now. I'll carry `fdc83b85` over in a separate PR to `main` once this merges, otherwise the drift just reverses direction.

PR #545, opened 2026-08-05
Say why the SARIF step failed, instead of letting jq say it badly

ejanalysis:
Rescues a commit that was stranded when [#543](https://github.com/Public-Environmental-Data-Partners/EJAM/pull/543) merged.

I pushed `6891d0d3` to that branch to address Copilot's review note, but #543 had already been squash-merged a few minutes earlier, so the commit never became part of the pull request and isn't on `development`. This re-applies it on top of the merged state — no other changes.

## What it does

Copilot pointed out that `Run lintr` is `continue-on-error: true`, so it can finish green having written no `lintr-results.sarif` at all — an R error, a missing dependency — and the `jq` step then fails.

Worth being precise about the risk: the job **already failed** in that case before #543, one step later at `upload-sarif`. This isn't a newly introduced failure mode. What was bad is the *message* — a bare `jq: No such file or directory`, which points the reader at the wrong step.

So it now fails deliberately, naming the step that actually broke:

```bash
if [ ! -f lintr-results.sarif ]; then
  echo "::error::lintr produced no lintr-results.sarif - check the 'Run lintr' step above for the R error."
  exit 1
fi
```

Failing rather than skipping is the right call here: lintr having died is exactly what this workflow exists to report, and swallowing it would leave the previous analysis standing while the check showed green.

On Copilot's second point — jq orders strings above numbers, so `"abc" < 1` is false and a non-numeric column would have been left alone rather than erroring. No behaviour change, but I added `select(type == "number")` so the filter doesn't rely on a type it never checks.

## Verified

All three paths, against a real 91-result SARIF:

| case | result |
|---|---|
| file missing | exits 1 with the annotation |
| `startColumn: 0`, `endColumn: 0` | 0 out-of-range remaining, all 91 results preserved |
| `startColumn: "oops"` | passed through untouched, all 91 results preserved |

### DIFF
--- a/.github/workflows/lintr.yaml
+++ b/.github/workflows/lintr.yaml
@@ -81,10 +81,11 @@ jobs:
             echo "::error::lintr produced no lintr-results.sarif - check the 'Run lintr' step above for the R error."
             exit 1
           fi
-          # select(type == "number") so a non-numeric column, should one ever
-          # appear, is passed through untouched instead of being compared.
-          jq '(.. | objects | select(has("startColumn")) | .startColumn | select(type == "number")) |= (if . < 1 then 1 else . end)
-              | (.. | objects | select(has("endColumn"))   | .endColumn   | select(type == "number")) |= (if . < 1 then 1 else . end)' \
+          # The type check sits on the right of |= so the left stays a plain
+          # path expression. A non-numeric column, should one ever appear, is
+          # passed through untouched rather than compared against 1.
+          jq '(.. | objects | select(has("startColumn")) | .startColumn) |= (if type == "number" and . < 1 then 1 else . end)
+              | (.. | objects | select(has("endColumn"))   | .endColumn)   |= (if type == "number" and . < 1 then 1 else . end)' \
             lintr-results.sarif > lintr-results.fixed.sarif
           mv lintr-results.fixed.sarif lintr-results.sarif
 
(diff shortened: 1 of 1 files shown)


ejanalysis:
Adopted in `04cc92cd` — thanks, this is the better way to write it.

One correction to the premise, though, since it matters for how much weight to give the change: `select()` **is** path-capable in jq, so `|=` doesn't reject it. The previous form ran clean in both environments —

- locally on `jq-1.7.1-apple`, and
- on `ubuntu-latest` in [run 31058152647](https://github.com/Public-Environmental-Data-Partners/EJAM/actions/runs/31058152647), where `Fix out-of-range columns in the SARIF file` went green

Direct comparison, same input, both forms:

```
$ echo '{"a":{"startColumn":0},"b":{"startColumn":"oops"},"c":{"startColumn":5}}'

# select on LHS  -> {"a":{"startColumn":1},"b":{"startColumn":"oops"},"c":{"startColumn":5}}  exit=0
# type check RHS -> {"a":{"startColumn":1},"b":{"startColumn":"oops"},"c":{"startColumn":5}}  exit=0
```

So this isn't a bug fix — behaviour is identical. I've taken it anyway because keeping the left side a plain path is the conventional idiom and removes any jq-version question at zero cost.

Re-tested with the step extracted straight from the YAML, so what I ran is what CI runs:

| case | result |
|---|---|
| file missing | exits 1 with the annotation |
| `startColumn: 0`, `endColumn: 0` | 0 out-of-range remaining, all 91 results preserved |
| `startColumn: "oops"` | passed through untouched, all 91 results preserved |

PR #551, opened 2026-08-08
Fix the R CMD check shinytest2 ERROR and three NOTEs

ejanalysis:
Takes `R CMD check` from **1 ERROR, 4 NOTEs** toward clean. Addresses the ERROR and three of the four NOTEs.

## The ERROR: shinytest2 — and a workflow that was testing nothing

`R CMD check` ran the full browser-driven app suite and died on *"An error occurred while waiting for Shiny to be stable"* — a `wait_for_idle(timeout = 5 * 1000)` that a cold check machine cannot meet. EJAM's wrapper had no CRAN or CI gate, only `skip_if_not_installed()`.

It now skips unless `SHINYTEST2_APP_DRIVER_TEST_ON_CRAN` is set — shinytest2's own opt-in variable, which `test-webapp-functionality.yaml` already sets.

**That alone would have traded a red check for no coverage at all**, because that workflow was not running these tests either. Every step filters to a `test-webapp-<category>-functionality.R` file; all of those call `shinytest2_webapp_functionality_individual()`, which skips unless `EJAM_SHINYTEST2_INDIVIDUAL` is set — and nothing set it. Run [31042164366](https://github.com/Public-Environmental-Data-Partners/EJAM/actions/runs/31042164366) logged *"Individual web app category tests are skipped by default"* while reporting success. Nine steps, all no-ops.

So this also sets `EJAM_SHINYTEST2_INDIVIDUAL: true` in that workflow. **Expect it to get slower and possibly go red** — these will be running for real, likely for the first time in a while. That is the point: green-because-skipped was worse.

## NOTE: `:::` calls to the package's own namespace

Dropped `EJAM:::` from `acs_endyear()`, `.arrow_ds_names`, and `localtree_exists()` in package code. Also moved an `@examples` line in `datasets_arrow_publish.R` inside its `\dontrun{}` — it referenced `EJAM:::.arrow_ds_names` and sat *outside* the block, so it was executing during checks.

## NOTE: no visible binding for global variable

Added the 24 data.table/dplyr NSE column names to the existing `globalVariables()` machinery, in their own vector so their provenance stays obvious.

**`stat_lorenz` is deliberately not declared.** It was in the same list, but it is a *function* from `gglorenz`, not a column — declaring it would have buried a genuinely undeclared dependency (the code comment even said *"not in Imports or Suggests for now"*). It is now `gglorenz::stat_lorenz()` with `gglorenz` added to Suggests, matching the fully-qualified style the rest of that draft file already uses.

## NOTE: Rd line widths

The 94-character flagged-areas heading was duplicated in two signatures and pushed `fill_tbl_full_subgroups.Rd`'s `\usage` past 90 characters. Now one internal constant used by both, with the literal recorded in `@param` so nothing is lost from the docs. Also split the over-long `url_ejamapi` example line.

Verified the constant is identical to the old literal and that **both** functions' defaults still resolve to it, so report output is unchanged.

## Not included

**NOTE: relative paths in package URLs** (~30 links across 11 vignettes). Left alone on purpose — those `../reference/index.html` links are *correct* on the pkgdown site and only look wrong to `R CMD check`, which inspects the built package. Fixing means either hardcoding the docs hostname, against the `url_package()` single-sourcing, or converting to `?topic` references that render differently on the site. Worth a decision, not a silent change.

**The four saved-results comparison failures are Windows-only and are not data drift** — see the PR discussion. Not touched here.

### DIFF
--- a/.github/workflows/test-webapp-functionality.yaml
+++ b/.github/workflows/test-webapp-functionality.yaml
@@ -37,6 +37,13 @@ jobs:
       GH_TOKEN: ${{ github.token }}
       NOT_CRAN: true
       SHINYTEST2_APP_DRIVER_TEST_ON_CRAN: 1
+      # Every step below filters to a test-webapp-<category>-functionality.R
+      # file, and those all call shinytest2_webapp_functionality_individual(),
+      # which skips unless this is set. It never was, so all nine steps were
+      # skipping and this workflow reported success while running no tests:
+      #   "Individual web app category tests are skipped by default."
+      # (verified in run 31042164366). Setting it makes them actually run.
+      EJAM_SHINYTEST2_INDIVIDUAL: true
       FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
     steps:
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -129,6 +129,7 @@ Suggests:
     datasets,
     devtools,
     fipio,
+    gglorenz,
     ggridges (>= 0.5.7),
     here,
     mapview,
--- a/R/aaa_globalVariables.R
+++ b/R/aaa_globalVariables.R
@@ -303,12 +303,20 @@ from_misc_add = sort(setdiff(from_misc, union(from_varlist_add, union(from_datas
 # here, rather than mixed into the lists above, so it stays obvious that they
 # came from a specific check run and can be re-derived from one.
 #
-# stat_lorenz was in this NOTE too but is deliberately NOT listed: it is a
-# function from gglorenz, not a column, so listing it would have hidden a
-# genuinely undeclared dependency. It is now called as gglorenz::stat_lorenz()
-# with gglorenz added to Suggests.
+# Two names from that NOTE are deliberately NOT listed, because neither is a
+# column and listing them would bury a real defect rather than quiet a false
+# positive:
+#
+#   stat_lorenz - a function from gglorenz, not a column. Now called as
+#     gglorenz::stat_lorenz() with gglorenz added to Suggests.
+#
+#   bysite - undefined inside plot_lorenz_distance_by_dcount(), which takes
+#     bybg_people and then pipes bysite, so it errors on the first call.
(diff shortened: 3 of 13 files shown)


ejanalysis:
Both correct, both pre-existing — and the second one caught a mistake in this PR.

## `bysite` — fixed here, in `adce9c41`

`plot_lorenz_distance_by_dcount()` takes `bybg_people` and then pipes `bysite`, which is defined nowhere in the function, so it errors on the first call. That name was in the "no visible binding" NOTE, and **I had added it to `globalVariables()` with the other 23** — which would have silenced the one automated signal pointing at the bug.

That is exactly the trap I called out and avoided for `stat_lorenz` two paragraphs earlier in the same commit, so it should have been avoided here too. `bysite` is now removed from the list, with the reasoning recorded next to it.

**Consequence, stated plainly:** the check will keep reporting a 1-name NOTE (`bysite`) until the draft file is repaired. That is the honest outcome — a clean check that hides a real defect is worse than an accurate NOTE.

## `filter()` on a literal and the constant `aes()` — not fixed here

Also right on both counts:

- `filter("Demog Index State Percentile" %in% c(...))` tests the *literal string*, which is not one of those two values, so it is `FALSE` and the filter drops every row. (`filter()` is unqualified too, unlike the `ggplot2::`/`scales::` prefixes used throughout the rest of the file.)
- `aes(colour = "Demog Index State Percentile")` maps a constant rather than the column created immediately above.

Not repaired in this PR on purpose. Both functions are unexported (`@noRd`), have apparently never run, and the file disagrees with itself about what it plots — `plot_lorenz_distance_by_dcount()` aggregates `pop * pctnhaa` while its axis label reads "Cumulative Percentage of Total Low Income Residents". Repairing that needs a decision about intent, not a guess, and this PR is about `R CMD check` hygiene.

Filed as #553 with all three defects and the `bysite`/NOTE relationship recorded, so the remaining NOTE has an obvious owner.

This PR still only touches that file to qualify `gglorenz::stat_lorenz()`.

Issue #553, opened 2026-08-10
plot_lorenz_*() draft plots are broken: undefined bysite, filter on a literal, constant aes mapping


ejanalysis:
Both draft Lorenz-curve plot functions in `R/plot_lorenz_popcount_by_site-DRAFT.R` have defects that would make them fail or produce a wrong plot on first call. Neither is exported (both are `@noRd` / `@keywords internal`), so nothing user-facing is affected.

Surfaced by Copilot review of #551, which touched that file only to qualify `gglorenz::stat_lorenz()`. All of these are pre-existing.

## 1. `plot_lorenz_distance_by_dcount()` — errors immediately

```r
plot_lorenz_distance_by_dcount <- function(bybg_people, varname = NULL) {
  ...
  bysite |>                       # <- bysite is never defined in this function
    ggplot2::ggplot(...)
```

The function takes `bybg_people` and then pipes `bysite`, which exists nowhere in scope. It errors on the first call. The plot caption also refers to `bysite`.

**This is why `bysite` is deliberately left out of the `globalVariables()` list in #551.** Declaring it would silence the "no visible binding" NOTE that is the only automated signal pointing at this bug. The check will keep reporting a 1-name NOTE until this is fixed — that NOTE is telling the truth.

## 2. `plot_lorenz_popcount_by_site()` — filter drops every row

```r
filter("Demog Index State Percentile" %in% c("High Demog.Index (at least 80th pctile in State)",
                                             "All Other Sites"))
```

This tests whether the *literal string* `"Demog Index State Percentile"` is one of those two values. It is not, so the condition is `FALSE` and the filter removes all rows. It needs a backticked column reference, and the column is created immediately above, so the filter may be redundant anyway.

`filter()` is also unqualified — `dplyr` is in Imports but the rest of this file consistently uses `ggplot2::` / `scales::` prefixes.

## 3. `plot_lorenz_popcount_by_site()` — constant colour mapping

```r
ggplot2::aes(x = pop, colour = "Demog Index State Percentile")
```

Maps a constant string rather than the column just created, so every point gets one colour and the legend shows the literal text. Needs backticks.

## Also worth a look

The two functions disagree with themselves about what is being plotted: `plot_lorenz_distance_by_dcount()` aggregates `pop * pctnhaa` while its axis label reads "Cumulative Percentage of Total Low Income Residents". Someone should decide what these are meant to show before repairing them.

## Suggested resolution

[...]

PR #554, opened 2026-08-10
plot_lorenz* not ready, so suppress R CMD NOTE for now

ejanalysis:
see #553 about how to fix these at some point

### DIFF
--- a/R/aaa_globalVariables.R
+++ b/R/aaa_globalVariables.R
@@ -307,14 +307,18 @@ from_misc_add = sort(setdiff(from_misc, union(from_varlist_add, union(from_datas
 # column and listing them would bury a real defect rather than quiet a false
 # positive:
 #
-#   stat_lorenz - a function from gglorenz, not a column. Now called as
-#     gglorenz::stat_lorenz() with gglorenz added to Suggests.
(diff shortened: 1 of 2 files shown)


ejanalysis:
Added `ea803ac8` here rather than as a separate PR, because the comment it fixes only becomes wrong when *this* PR merges — so the two changes belong together.

`R/aaa_globalVariables.R` carried a note next to `from_rcmdcheck_add` saying `bysite` *"errors on the first call"* and that leaving it undeclared *"keeps a 1-name NOTE that is telling the truth."* Both were accurate against `development`, and both stop being accurate the moment the draft bodies become comments here.

Reworded to past tense, and to say the thing that still matters: **keep `bysite` off that list**, or the bug comes back invisible the day someone uncomments that code. Also records that the `gglorenz` Suggests entry is temporarily unused for the same reason, so nobody prunes it as dead.

Comment only — no code change, and `bysite` is still not declared.

For the record on the PR itself: nothing calls either function and neither is exported (0 NAMESPACE entries), so the early `return(NULL)` has no blast radius. And the NOTE really does clear — the only remaining live mention of `bysite` is the parameter name in the signature, which is a defined binding, not an undefined global.

Issue #135, opened 2025-10-02
finish work in progress on default_ss_choose_method  vs input$ ?


ejanalysis:
see temp fix in v6 branch where tried to fix to use correct default/input/input but it did not work
so temporarily reverted to (confusingly/unintentionally?) using global_or_param() twice instead of using input$ for the non advanced tab ui default
         default_upload_dropdown = "upload", default_ss_choose_method = "upload", ## fix code in ui so 1 of these works

ejanalysis:
Last ranked: 2026-05-30
Cost score: 4
Benefit score: 11
Quadrant: B
Run report: inst/issues/ejam_issues_scored_by_risk_and_value.MD

ejanalysis:
may be fixed as of v3.2022.1 or v3.2022.2 -- check if can close

Issue #56, opened 2025-09-11
check/ fix Distance and Site Count summary stats, in doaggregate()


carlyhagen:
---ISSUE FROM ARCHIVE---
Originally created: 11/21/2023
-----------------

---ARCHIVED COMMENTS---

@ejanalysis is there any action needed on this ticket or can we close?|@ejanalysis is there any action needed on this ticket or can we close this one?|> @ejanalysis is there any action needed on this ticket or can we close this one?

leave it open - have not had time to look back at it. 

IMPORTANT: as this issue has be reuploaded from an archived repository, some of the links may be broken or no longer relevant. Critical resources can be found through https://ejanalysis.com/

ejanalysis:
Last ranked: 2026-05-30
Cost score: 6
Benefit score: 9
Quadrant: B
Run report: inst/issues/ejam_issues_scored_by_risk_and_value.MD

ejanalysis:
Triage note: this overlaps with #31 because both issues touch distance-related summary calculations in doaggregate(). They should probably be reviewed together.

Issue #31, opened 2025-09-11
fix distance_avg in doaggregate()$results_bybg_people


carlyhagen:
---ISSUE FROM ARCHIVE---
Originally created: 3/15/2023
-----------------

approx line 290 of doaggregate() should find avg distance of all people in a blockgroup to a single site, but $results_bybg_people$distance_avg seems to have the same distance_avg for a blockgroup to site 1 and same bg to site 2 -- they should differ. We want to have distance_by_group() be able to use the shorter distance to any of the nearby sites, for a given blockgroup but it somehow is already averaged across all sites the bg is near which isnt what we wanted and isnt what the code below should be doing. 

in doaggregate() 

sites2bgs_bysite    <-    sites2blocks[         , .(bgwt = sum(blockwt, na.rm = TRUE), 

                                                      proximityscore = stats::weighted.mean(proximityscore,   w = blockwt, na.rm = TRUE),

                                                      distance_avg   = stats::weighted.mean(distance,         w = blockwt, na.rm = TRUE),

                                                      sitedistance_min = min(sitedistance_min, na.rm = TRUE),

                                                      sitecount_max    = max(sitecount,        na.rm = TRUE)

  ), by=.(siteid, bgid)]

---ARCHIVED COMMENTS---

@ejanalysis can this be closed?|@ejanalysis what is the status of this ticket?|@DougMoy please check if this is still an issue.

IMPORTANT: as this issue has be reuploaded from an archived repository, some of the links may be broken or no longer relevant. Critical resources can be found through https://ejanalysis.com/

ejanalysis:
Last ranked: 2026-05-30
Cost score: 4
Benefit score: 9
Quadrant: B
Run report: inst/issues/ejam_issues_scored_by_risk_and_value.MD

ejanalysis:
Triage note: this overlaps with #56 because both issues touch distance-related summary calculations in doaggregate(). They should probably be reviewed together.

PR #562, opened 2026-08-11
chore: refresh open issue ranking report

ejanalysis:
Refreshes the live GitHub-derived open-issues prioritization report for 2026-08-11.

- 136 open issues scored
- Retains the capped 20-item Quadrant A focus shortlist
- Regenerates the Markdown, JSON, and HTML artifacts
- Does not modify GitHub rank labels

### DIFF
--- a/inst/issues/ejam_issues_scored_by_risk_and_value.MD
+++ b/inst/issues/ejam_issues_scored_by_risk_and_value.MD
@@ -1,6 +1,6 @@
 # EJAM Open Issues — Scored by Cost & Benefit
 
-**Total open issues:** 132  |  **Generated:** 2026-08-02  |  **Source:** Live GitHub REST API open issues (Public-Environmental-Data-Partners/EJAM)
+**Total open issues:** 136  |  **Generated:** 2026-08-11  |  **Source:** Live GitHub REST API open issues (Public-Environmental-Data-Partners/EJAM)
 
 **Score payload:** `inst/issues/ejam_issues_scored_by_risk_and_value.json`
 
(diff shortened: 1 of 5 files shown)


ejanalysis:
Confirmed, and the impact is larger than "malformed output" — **three issues were missing from the rendered report entirely**. Fixed in 91963e36.

An unescaped `$` opens a Pandoc inline-math span, and the span runs until the *next* `$`, swallowing everything between. Two titles did it — `fix distance_avg in doaggregate()$results_bybg_people` (#31) and `finish work in progress on default_ss_choose_method vs input$ ?` (#135) — and between them they consumed issues #553, #56 and #135 out of **both** the Quadrant B table and the Full Table:

| issue | links in HTML before | after |
|---|---|---|
| #31 | 3 | 3 |
| #553 | **1** (bullet only) | 3 |
| #56 | **1** | 3 |
| #135 | **1** | 3 |

42 `<td>` cells came back. `math inline` occurrences: 2 → 0.

## Fixed in the generator, not just the artifact

Escaping the `.MD` alone would have been undone by the next refresh, so the fix is a new `md_escape()` in `ejam_issues_scoring_functions.py`, applied at the three places a title is written into Markdown — the quadrant bullet list, the quadrant table, and the full table. The JSON payload keeps raw titles, since that is data rather than markup.

It escapes `|` too, which would end a table cell early. No current title contains one — I checked all 136 in the payload — but it is the same class of corruption and costs nothing to prevent.

Escaping happens *after* truncation, so a backslash cannot eat into the 70/72-character limit or leave a `\$` cut in half.

## Artifacts

The committed `.MD` is escaped to match exactly what the generator now emits (6 `$`, across the 3 sections; with 0 pipes in any title, nothing else differs). The `.html` was regenerated with the repo's own pipeline, `inst/issues/render_md_to_html.py`, rather than a hand-rolled pandoc call — all 4 `issue-details` tables still get their column layout.

## Tests

Two added to `test_ejam_issues_scoring_functions.py`, covering both `$` and `|` and asserting the escaping is applied in all three sections. I verified they are not vacuous by reducing `md_escape()` to the identity function and watching them fail. 8 pass.

Issue #159, opened 2025-10-02
integrate file-naming function with header-creating function


ejanalysis:
enhancement / refactoring - supports app and may fix bugs/issues in popups or report notes or header

integrate file naming function with header creating function:
```
  residents_within_xyz <- report_residents_within_xyz(
    sitetype = sitetype,
    radius = rad,
    nsites = nsites,  # but should note these are only the ones where $results_bysite$valid %in% TRUE
    area_in_square_miles = area_in_square_miles,
    sitenumber = sitenumber,
    ejam_uniq_id = ejamout1$ejam_uniq_id
  )

```
and
`create_filename()
`

ejanalysis:
Overlapping functions were written for making text to show in report header & popups & filename.
These should be merged, taking best of each approach:

- `report_residents_within_xyz()` and its helper `sitetype2text()`
- `buffer_desc_from_sitetype()` and its helper `site_method2text()`
- `create_filename()`

ejanalysis:
Last ranked: 2026-05-30
Cost score: 3
Benefit score: 7
Quadrant: A
Run report: inst/issues/ejam_issues_scored_by_risk_and_value.MD

PR #563, opened 2026-08-11
Say how sites were selected in the Excel notes tab; fix SIC/MACT and lowercase fips

ejanalysis:
Three related fixes to how EJAM says *how the sites were selected*. Started as the one-line `tolower()` fix below; grew after that fix turned out to reach nothing, which is now also fixed.

Rebased onto `development` after #483 merged.

## 1. `site_method2text()` — SIC and MACT branches were unreachable

The helper lowercases its input up front, then every branch compares against `tolower("...")` — except two, which compared against the uppercase literals directly:

```r
if (site_method %in% "SIC")  { return("EPA-regulated facilities by SIC code (industry type)") }
if (site_method %in% "MACT") { return("EPA-regulated facilities by MACT category (air toxics emissions source type)") }
```

Neither can ever match, so SIC and MACT fell through to `return("")`. Pre-existing on `development`; not introduced by #483, whose ZIP branch already uses `tolower()`. Found via a Copilot review comment on #483.

| site_method | before | after |
|---|---|---|
| `SIC` / `sic` | `""` | EPA-regulated facilities by SIC code (industry type) |
| `MACT` / `mact` | `""` | EPA-regulated facilities by MACT category (air toxics emissions source type) |

The other nine documented values are unchanged.

## 2. `buffer_desc_from_sitetype()` — the gate that never opened

Fixing #1 alone changed nothing users see, because the only caller never reached it:

```r
  if (buffer_desc == "") {                      # <- never TRUE
    based_on_txt <- site_method2text(site_method)
    if (based_on_txt %in% "")                   # <- inverted: appends only when empty
      buffer_desc <- paste0(buffer_desc, ", based on ", site_method2text(site_method))
  }
```

Every branch above assigns a non-empty string, so `buffer_desc == ""` is never `TRUE`. And the inner test is inverted — it would append `", based on <text>"` only when `<text>` was empty, producing a dangling `", based on "`. Both fixed.

This is what puts the text in front of a user, in the **notes tab of the Excel workbook** via `table_xls_from_ejam()`. (Report *headers* come from `sitetype2text()`, a different helper — see below.)

**Before → after, notes tab:**

| sitetype + site_method | before | after |
|---|---|---|
| `latlon` + `SIC` | Locations defined by latitude, longitude and radius | …, **based on EPA-regulated facilities by SIC code (industry type)** |
| `latlon` + `MACT` | Locations defined by latitude, longitude and radius | …, **based on EPA-regulated facilities by MACT category (air toxics emissions source type)** |
| `latlon` + `FRS` | Locations defined by latitude, longitude and radius | …, **based on EPA-regulated facilities by Registry ID** |
| `latlon` + `NAICS` | Locations defined by latitude, longitude and radius | …, **based on EPA-regulated facilities by NAICS code (industry type)** |
| `latlon` + `EPA_PROGRAM` | Locations defined by latitude, longitude and radius | …, **based on EPA-regulated Facilities by EPA program** |
| `shp` + `ZIP`/`zip`/`ZCTA` | Polygons defined by shapefile | …, **based on zip codes (ZCTA boundaries)** |
| `fips` + `FIPS_PLACE` | Census Units defined by FIPS code | …, **based on names of places** |
| `shp` + `SHP` | Polygons defined by shapefile | *unchanged — see below* |
| `fips` + `FIPS` | Census Units defined by FIPS code | *unchanged* |
| `latlon` + `latlon`/`mapclick` | Locations defined by latitude, longitude and radius | *unchanged* |

### Why the last three are deliberately left alone

The append is skipped when `site_method` only restates `sitetype`. That is the **common path, not a corner case** — `table_xls_from_ejam()` defaults `site_method` to `sitetype` when it is not supplied:

```r
  if (missing(site_method) || is.null(site_method) || site_method %in% "") {
    site_method <- sitetype
    if (site_method == 'shp' ) site_method <- 'SHP'
    if (site_method == 'fips') site_method <- 'FIPS'
  }
```

Without the skip, every plain shapefile analysis would read *"Polygons defined by shapefile, based on shapefile"*, and every FIPS one *"Census Units defined by FIPS code, based on FIPS codes"*.

### Edge cases

Both roxygen `@examples` of this function pass `site_method` **missing** (`EJAM:::buffer_desc_from_sitetype("shp")`), and the old dead gate meant that never mattered. The new code guards missing / `NULL` / `NA` / `""` / `character(0)` / unknown method and returns the plain description rather than erroring. Covered by tests.

## 3. `ejam2report()` — the FIPS gates were case-sensitive

The two gates that rebuild FIPS polygons used `site_method %in% "FIPS"`, so `site_method = "fips"` never fetched boundaries and the report came out unmapped. Now `toupper()`.

This matches what #483 already did for its ZIP gates 10 lines away, for exactly the same stated reason — and left the FIPS gates as the odd ones out:

```r
## toupper() and "ZCTA" accepted, because site_method2text() already treats both
## spellings and both cases as zip codes, so a caller passing site_method = "zip"
## should still get polygons. ejamit() itself always sets "ZIP".
if (toupper(site_method) %in% c("ZIP", "ZCTA") && is.null(shp) && ...
```

`toupper()` does not change the `NULL` behaviour: `NULL %in% "FIPS"` and `toupper(NULL) %in% "FIPS"` are both `logical(0)`. There is a pre-existing hole at [ejam2report.R:406-421](https://github.com/Public-Environmental-Data-Partners/EJAM/blob/development/R/ejam2report.R#L406-L421) where `site_method` can stay `NULL` if `ejamitout$site_method` is `NULL` and `sitetype` is none of shp/fips/latlon — unchanged by this PR, and it affects the existing ZIP gates identically.

## `sitetype2text()` — checked, not affected

It lowercases both `sitetype` and `site_method` and compares against lowercase literals throughout, so it has no such mismatch. Added assertions anyway so it stays that way — those four pass against the unpatched code, which is what confirms the difference.

## Tests

Added to **existing** files, so the group lists in `R/test_ejam.R` need no change (no auto-discovery there):

**`tests/testthat/test-sitetype2text.R`**
- `site_method2text()` returns the SIC/MACT text for both cases; no documented `site_method` falls through to `""`; vectorized input works; unknown still returns `""`
- `buffer_desc_from_sitetype()` appends the detail for SIC, MACT, FRS, NAICS, ZIP, FIPS_PLACE
- …and omits it for `shp`+`SHP`, `fips`+`FIPS`, `latlon`+`latlon`, `latlon`+`mapclick`
- …and handles missing / `NULL` / `NA` / `""` / `character(0)` / unknown without erroring, including both documented `@examples`
- `sitetype2text()` handles `SIC`/`sic`/`MACT`/`mact`

**`tests/testthat/test-ejam2report.R`**
- `ejam2report()` rebuilds FIPS polygons for `FIPS`, `fips`, and `Fips`, and does **not** for `NAICS`. Mirrors the mocked zip gate test #483 added in `test-shapes_from_zip.R`, including its check that the report itself still completes — otherwise "`shapes_from_fips()` was called" would pass even if everything after it blew up.

**Verified both directions.** With the fixes:

| file | result |
|---|---|
| `test-sitetype2text.R` | `FAIL 0 \| PASS 39` |
| `test-ejam2report.R` | `FAIL 0 \| SKIP 1 \| PASS 91` |
| `test-shapes_from_zip.R` | `FAIL 0 \| PASS 40` |
| `test-report_residents_within_xyz.R` | `FAIL 0 \| PASS 64` |
| `test-ejam2excel.R` | `FAIL 0 \| PASS 21` |

With the `R/` changes reverted: `test-sitetype2text.R` → `FAIL 8`, `test-ejam2report.R` → `FAIL 4`. So the new tests do fail without the fixes.

## `NEWS.md`

Entries added under the existing **EJAM v4 (unreleased)** section, in a new `## Bug Fixes` subsection, alongside #483's feature entry.

## Issue refs

No open issue covers these bugs, so there is nothing for the eventual `development` → `main` sync PR to restate as `Closes #N`. Related but not closed by this: #159 (reconcile `sitetype2text()` and `site_method2text()`, which both carry an in-code `*** per issue #159` note) and #483 (where the Copilot review surfaced the first one).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,28 @@ Features held for the v4 milestone, not part of the v3.2022.x patch line.
   maps their boundaries without needing the `shp` parameter. This automates the
   workflow that the Zipcodes article documented as manual steps (#482).
 
+## Bug Fixes
+
+- The notes tab of the Excel workbook now says how the sites were selected. It
+  had never done so: `buffer_desc_from_sitetype()` only appended that detail when
+  the description so far was empty, which none of its branches can produce, and
+  the test inside was inverted as well. A SIC analysis of latitude/longitude
+  sites now reads "Locations defined by latitude, longitude and radius, based on
+  EPA-regulated facilities by SIC code (industry type)" instead of stopping at
+  the radius. The detail is left off when it would only restate the site type,
+  so plain shapefile analyses do not read "Polygons defined by shapefile, based
+  on shapefile".
+
+- SIC and MACT analyses get their descriptions back. `site_method2text()`
+  lowercases its input, but its SIC and MACT branches compared against the
+  uppercase spellings, so neither could ever match and both fell through to an
+  empty string.
+
+- `ejam2report()` now fetches FIPS boundaries when `site_method` is given as
+  "fips" rather than "FIPS". The two gates that rebuild those polygons were
+  case-sensitive, so a lowercase spelling silently produced an unmapped report.
+  Matches how the zip code gates added for #482 already behave.
+
 
 # EJAM 3.2022.2 (August 2026)
 
--- a/R/community_report_helper_funs.R
+++ b/R/community_report_helper_funs.R
@@ -1251,10 +1251,10 @@ site_method2text =  function(site_method) {
     if (site_method %in% tolower("EPA_PROGRAM")) {
       return("EPA-regulated Facilities by EPA program")
     }
-    if (site_method %in% "SIC") {
+    if (site_method %in% tolower("SIC")) {
       return("EPA-regulated facilities by SIC code (industry type)")
     }
-    if (site_method %in% "MACT") {
+    if (site_method %in% tolower("MACT")) {
       return("EPA-regulated facilities by MACT category (air toxics emissions source type)")
     }
     return("")
--- a/R/community_report_helper_funs.R
+++ b/R/community_report_helper_funs.R
@@ -1200,10 +1200,30 @@ buffer_desc_from_sitetype <- function(sitetype, site_method) {
         } else {
           buffer_desc <- "Selected locations"
         }}}}
-  if (buffer_desc == "") {
-    based_on_txt <- site_method2text(site_method)
-    if (based_on_txt %in% "")
-      buffer_desc <- paste0(buffer_desc, ", based on ", site_method2text(site_method))
+  ## site_method can say more than sitetype does -- sitetype "latlon" with
+  ## site_method "NAICS" means points that are EPA-regulated facilities picked by
+  ## industry code, and sitetype "shp" with "ZIP" means zip code (ZCTA) polygons.
+  ## Append that detail when there is any to add.
+  ##
+  ## This used to be gated on buffer_desc == "", which no branch above can produce,
+  ## so it never ran; and the inner test was inverted, appending only when the text
+  ## was empty. Both are fixed here.
+  based_on_txt <- ""
+  if (!missing(site_method) && !is.null(site_method) && length(site_method) > 0 &&
+      !is.na(site_method[1])) {
+    ## Skip when site_method only restates sitetype, which would read
+    ## "Polygons defined by shapefile, based on shapefile". That is the common
+    ## case, not a corner one: table_xls_from_ejam() defaults site_method to
+    ## sitetype ('shp' -> 'SHP', 'fips' -> 'FIPS') whenever it is not supplied.
+    sitetype_known <- !missing(sitetype) && !is.null(sitetype) && length(sitetype) > 0
+    restates_sitetype <- sitetype_known &&
+      tolower(site_method[1]) %in% c(tolower(sitetype[1]), "mapclick")
+    if (!restates_sitetype) {
+      based_on_txt <- site_method2text(site_method[1])
+    }
+  }
+  if (!(based_on_txt %in% "")) {
+    buffer_desc <- paste0(buffer_desc, ", based on ", based_on_txt)
   }
   return(buffer_desc)
 }
--- a/R/community_report_helper_funs.R
+++ b/R/community_report_helper_funs.R
@@ -1186,7 +1186,12 @@ buffer_desc_from_sitetype <- function(sitetype, site_method) {
   ### *** per issue #159 should be reconciled/merged with
   ### buffer_desc_from_sitetype() and its helper site_method2text()
 
-  if (missing(sitetype) || is.null(sitetype)) {
+  ## NA and length-0 are treated like missing/NULL, not passed into the if() chain
+  ## below: ejamit_sitetype_from_output() returns NA when it cannot tell the type
+  ## (see R/ejamit_sitetype_from_.R), and if (NA == "shp") stops with "missing value
+  ## where TRUE/FALSE needed" - so an unknown site type crashed the description
+  ## instead of falling back to the generic wording.
+  if (missing(sitetype) || is.null(sitetype) || length(sitetype) == 0 || is.na(sitetype[1])) {
     buffer_desc <- "Selected Locations"
   } else {
     if (sitetype == "shp") {
@@ -1215,7 +1220,10 @@ buffer_desc_from_sitetype <- function(sitetype, site_method) {
     ## "Polygons defined by shapefile, based on shapefile". That is the common
     ## case, not a corner one: table_xls_from_ejam() defaults site_method to
     ## sitetype ('shp' -> 'SHP', 'fips' -> 'FIPS') whenever it is not supplied.
-    sitetype_known <- !missing(sitetype) && !is.null(sitetype) && length(sitetype) > 0
+    ## !is.na() here too, so restates_sitetype cannot come out NA and make
+    ## if (!restates_sitetype) error the same way
+    sitetype_known <- !missing(sitetype) && !is.null(sitetype) &&
+      length(sitetype) > 0 && !is.na(sitetype[1])
     restates_sitetype <- sitetype_known &&
       tolower(site_method[1]) %in% c(tolower(sitetype[1]), "mapclick")
     if (!restates_sitetype) {
--- a/R/ejam2report.R
+++ b/R/ejam2report.R
@@ -482,7 +482,11 @@ ejam2report <- function(ejamitout = testoutput_ejamit_10pts_1miles,
     ## toupper() for the same reason as the zip gate just below: site_method2text()
     ## and sitetype2text() both accept either case, so a caller passing "fips"
     ## should still get polygons. ejamit() itself always sets "FIPS".
-    if (toupper(site_method) %in% "FIPS" && is.null(shp)) {
+    ## isTRUE(), because site_method can still be NULL here - the defaulting block
+    ## above leaves it unset when ejamitout$site_method is NULL and sitetype is
+    ## none of shp/fips/latlon - and %in% on NULL gives logical(0), which makes
+    ## if() error with "argument is of length zero" rather than skipping the gate.
+    if (isTRUE(toupper(site_method) %in% "FIPS") && is.null(shp)) {
       shp <- shapes_from_fips(ejamitout$results_bysite$ejam_uniq_id)
       if (!is.na(rad) && rad > 0 && rad != 999) {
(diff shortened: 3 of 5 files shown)


ejanalysis:
Worked through the three suppressed comments from the latest review. Two were real and are fixed in 94f02bfb; one no longer applies.

## Fixed — the "all documented site_method values" test did not cover them all

Correct. `site_method_options` listed nine values, but #483 added `ZIP` (and `ZCTA` as its synonym) to the `@param site_method` roxygen list, so the assertion was claiming more coverage than it had and a ZIP/ZCTA regression would have slipped through. Both added — all 11 now return non-empty text in either case.

## Fixed — the NAICS negative case was passing vacuously

This was the good catch, and it was worse than theoretical: the case really was passing for the wrong reason.

Asserting only that `shapes_from_fips()` was *not* called is equally satisfied by `ejam2report()` dying before it ever reaches the gate. I checked by adding `expect_false(inherits(res, "try-error"))` and watching it fail — with a non-FIPS method against this FIPS-shaped fixture, `shp` stays `NULL`, the unmapped branch runs, and it errors on missing `ratio.to.state.avg.*` columns that the minimal fixture omits.

So the case *cannot* assert completion — that would mean padding the fixture with a dozen ratio columns that have nothing to do with what this test is about. It now pins that specific downstream error instead, which is only reachable **past** the gate. That distinguishes a correct skip from an early crash, which is precisely what was missing.

## No longer applies — PR description vs. actual scope

The description now documents the `buffer_desc_from_sitetype()` change as its own numbered section, with a before/after table of the notes-tab wording and an explicit rationale for the three cases deliberately left unchanged. This comment appears to have been raised against an earlier revision of the body.

Tests: `test-sitetype2text.R` 39, `test-ejam2report.R` 94, `test-shapes_from_zip.R` 40, `test-ejam2excel.R` 21 — 0 failures, 0 warnings.

ejanalysis:
Confirmed and fixed in 1c93be6e — and the scope is wider than the two lines flagged.

Verified the crash directly on R 4.6.0:

```
site_method = NULL          -> ERROR: argument is of length zero
site_method = character(0)  -> ERROR: argument is of length zero
site_method = NA            -> FALSE branch (fine)
site_method = "FIPS"        -> TRUE branch (fine)
```

`%in%` on `NULL` gives `logical(0)`, and `if (logical(0) && ...)` errors rather than skipping the gate — so report generation dies outright instead of merely not rebuilding polygons.

## Why I fixed four gates, not two

The PR description is correct that `toupper()` does not introduce this, and that the hole at `ejam2report.R:406-421` predates it — `site_method` stays unset when `ejamitout$site_method` is `NULL` and `sitetype` is none of shp/fips/latlon, because that `else` chain has no final fallback.

But "pre-existing, therefore out of scope" understates it here: **the two ZIP gates added by #483 are new instances of the same hazard, and they are already merged to `development`.** So four gates were exposed, not two. Guarding only the FIPS pair would have left the ZIP pair to be rediscovered later — with the added irony that #483's stated reason for touching those lines was to stop FIPS and ZIP behaving differently.

All four now use `isTRUE(...)`. That rather than a length check, because it collapses `logical(0)` to `FALSE`, which is the behavior you want anyway: an unknown method should skip the rebuild, not crash. Confirmed it changes nothing else — with a `NULL` `site_method` the length-zero error is gone, and `"fips"` / `"zcta"` still match.

## On the test

It asserts that no *unguarded* gate remains, deliberately phrased as the absence of the bad pattern rather than a count of the good one, so legitimately adding or removing a gate does not break it.

Flagging one thing honestly: this is a source-level assertion, which is the pattern I argued against on #483 and replaced there with a behavioral test. The difference is cost — reaching each of these four gates behaviorally takes minutes per `ejam2report()` call against a fixture that then fails on unrelated missing columns, whereas #483's ZIP gate was reachable in milliseconds under mocks. If that trade seems wrong, the alternative is a slow integration test and I am happy to add one.

`test-ejam2report.R` 104, `test-sitetype2text.R` 39, `test-shapes_from_zip.R` 40 — 0 failures.

ejanalysis:
All three confirmed and fixed in b787456f.

## `buffer_desc_from_sitetype()` on an `NA` sitetype — reachable, not hypothetical

The pointer to `ejamit_sitetype_from_output()` is the important part: it returns `NA` when it cannot determine the type (`R/ejamit_sitetype_from_.R:128`), and that `NA` flows straight into `if (sitetype == "shp")`.

```
before:  buffer_desc_from_sitetype(NA_character_)         -> ERROR: missing value where TRUE/FALSE needed
         buffer_desc_from_sitetype(NA_character_, "SIC")  -> ERROR
after:   buffer_desc_from_sitetype(NA_character_)         -> "Selected Locations"
         buffer_desc_from_sitetype(NA_character_, "SIC")  -> "Selected Locations, based on EPA-regulated facilities by SIC code (industry type)"
```

`NA` and length-0 are now handled alongside missing/`NULL`, so an undetermined site type degrades to the generic wording instead of killing the description. Happy paths are untouched — `latlon` + `SIC` still reads "Locations defined by latitude, longitude and radius, based on EPA-regulated facilities by SIC code (industry type)".

This is the same failure family as the `NULL` `site_method` gates in the previous commit, which is a fair signal that this PR walked into a corner of the code where nothing guards its inputs.

## `restates_sitetype` could come out `NA`

Also correct, and it would have failed at `if (!restates_sitetype)` in exactly the same way. `sitetype_known` now includes `!is.na(sitetype[1])`, so the short-circuit keeps it strictly `TRUE`/`FALSE`.

## The source-inspection test — a real problem, and mine

This one matters more than it looks. `R/` is not shipped in an installed package, so `readLines()` would have **errored** during `R CMD check` rather than skipping — and `check()` is part of the release pre-merge routine, so this would have surfaced as a release-blocking failure well away from where it came from.

Now gated with `skip_if_not(file.exists(...))`, matching the source-inspection test in `test-MAP_FUNCTIONS.R:474-476` that you pointed at. The match is also whitespace-insensitive now (a regex on `if\s*\(\s*toupper\s*\(\s*site_method\s*\)\s*%in%`), so reformatting the condition cannot hide a bare gate.

Taken with the point I raised on the previous round, I'd read this as the source-inspection test having earned its keep only if it is cheap and robust — it is now both, but it stays the weakest assertion in this PR.

`test-ejam2report.R` 104, `test-sitetype2text.R` 39, `test-shapes_from_zip.R` 40, `test-ejam2excel.R` 21, `test-MAP_FUNCTIONS.R` 89 — 0 failures, 0 warnings.

ejanalysis (commented):
Reviewed current head `b787456f` of Public-Environmental-Data-Partners/EJAM#563. The two newly-open Copilot threads about the polygon guards do not apply: each condition is `isTRUE(toupper(site_method) %in% ...) && ...`, so `isTRUE()` receives only the `%in%` result and safely turns NULL/length-zero values into `FALSE`. I also confirmed the new NA/length-zero `sitetype` fallbacks in a detached worktree. No new actionable issue found.

— Codex

ejanalysis:
Checked this against the parse tree rather than the prose, and it does not apply — `isTRUE()` wraps only the `%in%`, and the `&&` sits outside it.

The line is:

```r
if (isTRUE(toupper(site_method) %in% "FIPS") && is.null(shp)) {
```

Asking R how it actually parses that:

```
outermost call: &&
isTRUE() arg  : toupper(site_method) %in% "FIPS"      <- only the %in%, no && inside
```

So `isTRUE()` receives `logical(0)` and returns `FALSE`, and `&&` then evaluates `FALSE && is.null(shp)` — which short-circuits without ever seeing a zero-length value. The comment's premise, that `&&` runs on the `%in%` result before `isTRUE()` can act, has the evaluation order backwards: `isTRUE()` is an argument to `&&`, so it is forced first.

Confirmed by running the exact expression:

```
site_method = NULL          -> FALSE-branch
site_method = character(0)  -> FALSE-branch
site_method = FIPS          -> TRUE-branch
site_method = fips          -> TRUE-branch
```

No error in any case, and the matching cases still match. This is also what the earlier end-to-end check showed: with a `NULL` `site_method`, `ejam2report()` no longer raises the length-zero error and instead proceeds to fail later on unrelated fixture columns — i.e. it got past these gates.

The suggested reorder ("so `isTRUE()` only guards the `%in%` check") is what the code already does, so there is nothing to change. Resolving both threads.

PR #568, opened 2026-08-11
Guard the report-description helpers against inputs they can actually get

ejanalysis:
Review of #563 turned up three latent crashes one at a time — a `NULL` `site_method`, an `NA` `sitetype`, a `logical(0)` in an `if()`. That pattern suggested the report-description helpers are generally unguarded rather than those three being unlucky, so instead of waiting for the next review to find the fourth, I probed them directly with `NULL`, `NA`, `character(0)`, `""` and unknown values.

Five more crashes, and one correctness bug the probing turned up on the way.

## The correctness bug — a zero-row latlon analysis was reported as FIPS

`all()` on zero elements is vacuously `TRUE`, so `all(fips_valid(character(0)))` returned `TRUE` and a 0-row table with an `ejam_uniq_id` column was labelled `"fips"` on the strength of no values at all.

This is reachable, not theoretical. Drop every row of a **latlon** analysis — which is what happens once invalid sites are removed, and is exactly the case covered by the existing "popup_from_ejscreen() handles zero-row results after invalid shapes are dropped" test — and you are left with `lat`/`lon` **and** `ejam_uniq_id` columns. The lat/lon branch correctly declines an all-NA table, and then the vacuous `all()` claimed FIPS:

```
before:  sitetype_from_dt(latlon_results[0, ]) -> "fips"
after:   sitetype_from_dt(latlon_results[0, ]) -> NA, with the existing
         "cannot determine valid sitetype" warning
```

## The crashes

| function | input | was | now |
|---|---|---|---|
| `sitetype_from_dt()` | `NULL` | **ERROR** argument is of length zero | `NA` (its own documented fallback) |
| `ejamit_sitetype_from_output()` | `NULL` | **ERROR** (via the above) | `NA` |
| `sitetype2text()` | `sitetype = character(0)` | **ERROR** missing value where TRUE/FALSE needed | `"place"` — same as `NULL` |
| `sitetype2text()` | `site_method = character(0)` | **ERROR** | same as `NULL` |
| `ejam2report()` | unrecognized `sitetype` | left `site_method` `NULL` | falls back to `""` |

`character(0)` is not `NULL`, so it slipped past the `is.null()` defaults, survived the `is.na()` replacement unchanged, and then `sitetype %in% ...` yielded `logical(0)`, which an `if()` cannot use.

The `ejam2report()` one is the root of the earlier symptoms: the defaulting block sets `site_method` from `ejamitout$site_method` or from `sitetype` shp/fips/latlon, with **no final else** — so an unrecognized sitetype, including the `NA` that `ejamit_sitetype_from_output()` returns when it cannot tell, left it unset and every downstream use had to defend itself separately. That is why the polygon-rebuild gates needed `isTRUE()`. It now falls back to `""`, which the block's own missing/blank test already treats as "not supplied".

## Deliberately not guarded

Passing a character vector where an `ejamit()` output list belongs still errors. That is a caller bug, and papering over it would hide the mistake — there is a test asserting it keeps erroring.

## One behavior change worth knowing about

The `"cannot determine valid sitetype"` warning is now emitted in one existing passing test, where previously the code silently mislabelled the empty result as FIPS. That is the fix working, but it means this warning will appear in logs wherever a zero-row result gets summarized. If that proves noisy, the alternative is to make the explicitly-empty case return `NA` silently and keep the warning only for "has rows but cannot classify" — say the word and I'll split it that way.

## Verification

12 test files, **0 failures**, and the single intended warning above: `MAP_FUNCTIONS` 90, `ejam2map` 41, `ejam2histogram` 18, `ejam2shapefile` 28, `ejam2excel` 21, `ejam2report` 115, `ejamit` 63, `ejamit_sitetype_from_input` 13, `ejamit_sitetype_from_output` 23, `sitetype2text` 46, `shapes_from_zip` 40, `shapefix` 23.

Tests added for every case in the table, for the zero-row latlon regression specifically, and for the wrong-class input continuing to error. No roxygen changed, so no `man/` regeneration; both test files were already registered in `R/test_ejam.R`.

Relates to Public-Environmental-Data-Partners/EJAM#159, which notes these helpers should be reconciled — this does not do that, only stops them crashing.

### DIFF
--- a/R/community_report_helper_funs.R
+++ b/R/community_report_helper_funs.R
@@ -1338,8 +1338,11 @@ sitetype2text <- function(sitetype = NULL, site_method = sitetype, sitetype_null
   if (is.null(nsites) || any(is.na(nsites))) {
     nsites <- 99 # just makes it plural, e.g., "places"
   }
-  if (is.null(sitetype))    {sitetype    <- sitetype_nullna}
-  if (is.null(site_method)) {site_method <- sitetype}
+  ## length-0 is handled alongside NULL, not left to fall through: character(0)
+  ## survives the is.na() replacement below unchanged, and then `sitetype %in% ...`
+  ## yields logical(0), which makes the if() below error instead of simply not matching.
+  if (is.null(sitetype)    || length(sitetype) == 0)    {sitetype    <- sitetype_nullna}
+  if (is.null(site_method) || length(site_method) == 0) {site_method <- sitetype}
 
   sitetype[   is.na(sitetype)]    <- sitetype_nullna
   site_method[is.na(site_method)] <- sitetype_nullna
--- a/R/ejam2report.R
+++ b/R/ejam2report.R
@@ -398,6 +398,17 @@ ejam2report <- function(ejamitout = testoutput_ejamit_10pts_1miles,
     ejamitout$sitetype <- ejamit_sitetype_from_output(ejamitout)
   }
   sitetype <- ejamitout$sitetype
+  ## Normalized once here, so none of the ~9 `sitetype %in% ...` tests below can see a
+  ## length-0 value: NULL %in% "shp" is logical(0), which if() cannot use. This is
+  ## reachable - an ejamitout carrying an explicit NULL sitetype still has "sitetype"
+  ## in names(), so the inference above is skipped and NULL arrives here.
+  ## NA is deliberately left as NA: %in% returns FALSE for it, never NA, so it is
+  ## already safe in every test below, and keeping it distinguishes "could not tell"
+  ## from "shp"/"fips"/"latlon".
+  if (is.null(sitetype) || length(sitetype) == 0) {
+    sitetype <- NA
+    ejamitout$sitetype <- sitetype # keep the two in step
+  }
 
   # 2d, get more detailed info about how site was specified, from "site_method" parameter,
   # which server stores as the submitted_upload_method() reactive
--- a/R/ejam2report.R
+++ b/R/ejam2report.R
@@ -415,6 +415,14 @@ ejam2report <- function(ejamitout = testoutput_ejamit_10pts_1miles,
       } else {
         if (sitetype %in% 'latlon') {
           site_method <- 'latlon'
+        } else {
+          ## final fallback, so site_method is always set from here on. Without it an
+          ## unrecognized sitetype (including the NA that ejamit_sitetype_from_output()
+          ## returns when it cannot tell) left site_method NULL, and every downstream
+          ## use had to defend itself separately - which is how the polygon-rebuild
+          ## gates came to need isTRUE(). "" is what the missing/blank test at the top
+          ## of this block already treats as "not supplied".
+          site_method <- ""
         }
       }
     }
--- a/R/ejam2report.R
+++ b/R/ejam2report.R
@@ -406,7 +406,7 @@ ejam2report <- function(ejamitout = testoutput_ejamit_10pts_1miles,
   ## already safe in every test below, and keeping it distinguishes "could not tell"
   ## from "shp"/"fips"/"latlon".
   if (is.null(sitetype) || length(sitetype) == 0) {
-    sitetype <- NA
+    sitetype <- NA_character_ # character, since sitetype is otherwise "latlon"/"fips"/"shp"
     ejamitout$sitetype <- sitetype # keep the two in step
   }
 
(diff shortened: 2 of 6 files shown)


ejanalysis:
Right line, wrong reason — and probing it turned up a real hole the comment missed. Fixed in f2ef9459.

## The stated mechanism does not hold

`%in%` never returns `NA`. It is `match(x, table, nomatch = 0) > 0`, and `match()` returns 0 for an `NA` that is not in the table, so the result is `FALSE`:

```
NA(logical)    %in% "shp"  -> FALSE    is.na = FALSE
NA_character_  %in% "shp"  -> FALSE    is.na = FALSE
```

So an `NA` `sitetype` already flows through the chain and lands on the new `site_method <- ""` fallback exactly as intended — I ran the chain to confirm it returns `""` for both `NA` and `NA_character_`. `isTRUE()` there would guard something that cannot occur.

Worth noting this is the second time this review has raised "`%in%` can return `NA`" — the same claim came up on Public-Environmental-Data-Partners/EJAM#558 against `st_geometry_type(shp) %in% c("POINT", "MULTIPOINT")` and was declined for the same reason.

## But there is a real bug at that line

Length-0 is a different matter, and it does crash:

```
NULL          %in% "shp"  -> logical(0)   -> if() errors, "argument is of length zero"
character(0)  %in% "shp"  -> logical(0)   -> if() errors
```

And it is reachable: an `ejamitout` carrying an **explicit `NULL`** `sitetype` still has `"sitetype"` in `names()`, so the inference at line 397 is skipped and `NULL` arrives at the chain directly.

Fixed by normalizing `sitetype` once, immediately after it is read, rather than wrapping each `%in%` — this function has about nine of them (also at ~524, 599, 602, 829, 839), and one normalization covers them all instead of leaving the later ones to be found separately. `ejamitout$sitetype` is kept in step with the local value.

`NA` is deliberately left as `NA`: it is already safe in every test, and keeping it preserves the distinction between "could not determine the type" and a real type — which is the whole point of `sitetype_from_dt()` returning `NA` in this PR.

Test covers `NULL`, `character(0)`, `NA`, `NA_character_` and an unknown string all reaching `""`, the recognized types still mapping to `SHP`/`FIPS`/`latlon`, and asserts both facts above directly: `%in%` yields `FALSE` (not `NA`) for `NA`, and length 0 for `NULL`.

`test-ejam2report.R` 116, `test-ejamit_sitetype_from_output.R` 23, `test-sitetype2text.R` 46, `test-ejam2excel.R` 21 — 0 failures.

ejanalysis:
Both correct, both on code this PR introduced. Fixed in 09ff6b93.

## `site_method = character(0)` — the same defect, in the one line I didn't touch

This is a fair hit. The whole point of this PR is that length-0 slips past `is.null()` and dies at `%in%`, and I added a fallback to the *bottom* of that block while leaving exactly that bug in its *opening condition*:

```
before:  site_method = character(0)  ->  ERROR: argument is of length zero
after:   site_method = character(0)  ->  enters defaulting, same as NULL and ""
```

Fixed with a `length()` test, plus `isTRUE(site_method[1] %in% "")` rather than a bare `%in%` — that also makes a length > 1 `site_method` well defined instead of leaving another multi-element `if()` condition waiting to error.

## `NA` vs `NA_character_` — agreed, and extended one step further

Right that a logical `NA` in a field that is otherwise `"latlon"`/`"fips"`/`"shp"` is inconsistent. I checked nothing compares it with `identical()` before changing it — only `is.na()`, which is unaffected either way.

Rather than only typing the normalization, I also changed the two `"cannot determine valid sitetype"` returns in `sitetype_from_dt()` to `NA_character_`. Fixing just the one you flagged would have left a logical `NA` coming out of `sitetype_from_dt()` and a character `NA` coming out of the normalization — two spellings of the same "cannot tell", which is how this class of inconsistency starts. Now there is one.

Worth flagging that this is the **opposite** call from the `NA_character_` suggestion I declined on Public-Environmental-Data-Partners/EJAM#558, so the reasoning should be legible rather than looking arbitrary. There, the change would have flipped an entire column from logical to character on *every valid upload* — a happy-path behavior change riding along on an unrelated fix. Here it types a scalar that is already conceptually character, only in the degenerate case, in a PR whose subject is exactly these degenerate cases.

## Verification

`test-ejam2report.R` 121, `test-ejamit_sitetype_from_output.R` 23, `test-sitetype2text.R` 46, `test-ejam2excel.R` 21, `test-MAP_FUNCTIONS.R` 90, `test-ejamit.R` 63 — 0 failures, plus the single intended zero-row warning described in the PR body.

Tests extended to cover the `character(0)` `site_method`, the `""` and `NULL` cases still opening the defaulting, and a length > 1 value being well defined rather than an error.

